PR #5, opened 2021-05-21
Adding cluster cells the same way as #426 in FCCSW

kjvbrt:
Hello,

This request adds ability to add cluster cells the same way as #426 in FCCSW.
Needs change of ITowerTool interface in k4FWCore.

Best,
Juraj


vvolkl:
Thanks @kjvbrt ! Would it be possible to use edm4hep CalorimeterHits here too?

vvolkl (approved):

BrieucF:
I think this last commit is going to break ReDoSegmentation, I just changed it so that it is fed with SimCalorimeterHits https://github.com/HEP-FCC/k4SimGeant4/pull/6

vvolkl:
Tested locally that this works with https://github.com/key4hep/k4FWCore/pull/45

PR #11, opened 2021-06-14
Fix cluster duplicates

BrieucF:
This prevents to store, in the new event, the clusters from the previous events. Not sure why/if it was working in the previous fccedm implementation...


vvolkl:
Ah. This was working before due to this line, which was just removed in the last PR: https://github.com/HEP-FCC/k4RecCalorimeter/commit/ab0a4a386e3cdd61b0d81a8b81e662b99a80c5e1#diff-66c96baafdc27cd175cac456d156562324919198f6f10793058acb4c7292c39fL61


Redeclaring something with an `m_` prefix as a local variable is of course not proper style and very confusing, much better to use `clear()` like you did. Thanks! We should also add a test where this error would be caught.

PR #13, opened 2021-07-19
Fix the (a?) calo clustering segmentation fault

BrieucF:
This line was creating a segmentation fault:
https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecCalorimeter/src/components/CaloTowerTool.cpp#L197
Both clears seem to be necessary to avoid the problem, not exactly sure why... Explanations and opinions on how to best fix this would be very welcome!


BrieucF:
This actually breaks the functionality of attaching cells to the cluster in case we have two clusters in the event

PR #15, opened 2021-10-04
[LAr] new sampling fractions

BrieucF:


BrieucF:
Goes along with https://github.com/HEP-FCC/FCCDetectors/pull/14

PR #17, opened 2021-10-18
Fixes for tutorials

vvolkl:


RecCalorimeter/src/components/CaloTowerTool.cpp
```diff
@@ -100,7 +100,16 @@ StatusCode CaloTowerTool::initialize() {
   return StatusCode::SUCCESS;
 }
 
-StatusCode CaloTowerTool::finalize() { return GaudiTool::finalize(); }
+StatusCode CaloTowerTool::finalize() { 
+  
+  
+for (auto& towerInMap : m_cellsInTowers) {
```
BrieucF, line 106: Hi Valentin, great! Do you think it will fix this as well: https://github.com/HEP-FCC/k4RecCalorimeter/pull/13 ?

PR #21, opened 2022-02-02
Adapt to new podio generated class names

vvolkl:
- Replace some dedicated types with `auto` and rename others (Cluster -> MutableCluster, ConstCluster -> Cluster) to adapt to the new naming scheme introduced in AIDASoft/podio#205.
- The necessary changes in EDM4hep: key4hep/EDM4hep#132 would break this otherwise. Due to the explict use of Mutable* (in attachCells) this now requires a new version of edm4hep.
- Due to the change in the interface of attachClusters this also now requires a new version of k4fwcore.

### DIFF
--- a/RecCalorimeter/src/components/CorrectECalBarrelSliWinCluster.cpp
+++ b/RecCalorimeter/src/components/CorrectECalBarrelSliWinCluster.cpp
@@ -254,7 +254,7 @@ StatusCode CorrectECalBarrelSliWinCluster::execute() {
     }
 
     // 0. Create new cluster, copy information from input
-    edm4hep::Cluster newCluster = correctedClusters->create();
+    edm4hep::MutableCluster newCluster = correctedClusters->create();
     double energy = 0;
     newCluster.setPosition(cluster.getPosition());
     for (auto cell = cluster.hits_begin(); cell != cluster.hits_end(); cell++) {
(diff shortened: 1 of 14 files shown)


RecCalorimeter/src/components/CorrectECalBarrelSliWinCluster.cpp
```diff
@@ -254,7 +254,7 @@ StatusCode CorrectECalBarrelSliWinCluster::execute() {
     }
 
     // 0. Create new cluster, copy information from input
-    edm4hep::Cluster newCluster = correctedClusters->create();
+    edm4hep::MutableCluster newCluster = correctedClusters->create();
```
BrieucF, line 257: Can't we use 'auto' there as well?

BrieucF (commented):
Hi Valentin, thanks! Could it be tested?

RecCalorimeter/src/components/CorrectECalBarrelSliWinCluster.cpp
vvolkl: Yes indeed!

```suggestion
    auto newCluster = correctedClusters->create();
```

vvolkl:
Thanks for taking a look, the best way to test it is probably to merge it and then use the nightly builds.

BrieucF:
Alright, thanks!

PR #28, opened 2022-07-05
Changes for speed improvements in clustering jobs

gartrog:
This PR should be merged at the same time as the one in k4FWCore.

Lots of small things that when put together allow to gain a lot.

- CaloTopoClusters: use unordered_map instead of map
- CaloTopoClusters: fix compilation warning
- CaloTopoClusterInputTool: optimize use of map: update values instead
  of clearing and filling again
- CaloTopoClusterInputTool: remove unused member variable
- CaloTowerTool: update to use the new interface (do not fill map of
  links if not needed)
- CaloTowerTool: micro-optimisation, reduce calls to eta and phi.
- CreateCaloClustersSlidingWindow: use new interface of CaloTowerTool to
  not fill map of links if attachCells is not used
- NoiseCaloCellsFromFileTool: use Decoder::get(ID, int) instead of string
- NoiseCaloCellsFromFileTool: simplify finding bin in histogram
- NoiseCaloCellsFromFileTool: micro-optimisation: avoid sqrt(x*x) when
  no pileup noise.
- SplitClusters: fix dangerous compilation warning
- CaloTopoClusterFCCee: use Decoder::get(ID, int) instead of string
- CaloTopoClusterFCCee: to avoid too many lookups in the huge noise map,
  precompute minimum noise per layer and start by comparing noise to
  this minimum noise before looking into the big map.
- CaloTopoClusterFCCee: micro-optimisation: avoid instantiating a
  vector<vector<pair>> of size 100000 at every call...
- CaloTopoClusterFCCee: fix compilation warning
- CreateCaloCellPositionsFCCee: create and use a cache of cell positions
  to avoid geomtry calls and conversions at every event
- CreateCaloCellPositionsFCCee: use Decoder::get(ID, int) instead of string

Note: many alogirithmic optimisations are certainly still possible, I
focused on easy gains that did not require deep understanding of the
code.
I suspect many calls related to geometry are far from optimal.

### DIFF
--- a/RecCalorimeter/src/components/CaloTopoCluster.cpp
+++ b/RecCalorimeter/src/components/CaloTopoCluster.cpp
@@ -80,7 +80,7 @@ StatusCode CaloTopoCluster::initialize() {
 
 StatusCode CaloTopoCluster::execute() {
   
-  std::map<uint64_t, double> allCells;
+  std::unordered_map<uint64_t, double> allCells;
   std::vector<std::pair<uint64_t, double>> firstSeeds;
   
   // get input cell map from input tool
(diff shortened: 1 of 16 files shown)


gartrog:
Tagging @vvolkl and @BrieucF

BrieucF:
Just tested it and everything looks ok to me!

PR #34, opened 2023-02-10
fix for running standalone HCal

mmlynari:
fix for running standalone HCal, otherwise, the code always looks for ECal

### DIFF
--- a/RecCalorimeter/src/components/CaloTowerTool.cpp
+++ b/RecCalorimeter/src/components/CaloTowerTool.cpp
@@ -48,7 +48,9 @@ StatusCode CaloTowerTool::initialize() {
     error() << "Wrong type of segmentation" << endmsg;
     return StatusCode::FAILURE;
   }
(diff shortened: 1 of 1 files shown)


kjvbrt:
What is the intended use of the `m_useHalfTower` variable?

Is it to ignore cells from the layers above 3? [link](https://github.com/HEP-FCC/k4RecCalorimeter/blob/688bb6a2418416c54a7347c464fc9bca10029d6f/RecCalorimeter/src/components/CaloTowerTool.cpp#L337)

But the code will ignore all cells after first cell with layerID larger than 3 (including the current cell).

Also, 3 is a magic number. Better to have it as a configurable value.

mmlynari:
my understanding was that this was an option that was used for some FCC-hh studies. I do not think we are currently using it. Maybe @faltovaj or @BrieucF can comment if this option can be of any use in the future

BrieucF:
I'll merge and fix `m_useHalfTower` in another PR

PR #38, opened 2023-03-14
add benchmark method to CorrectCaloClusters

mmlynari:
add benchmark method to CorrectCaloClusters

### DIFF
--- a/RecCalorimeter/src/components/CorrectCaloClusters.cpp
+++ b/RecCalorimeter/src/components/CorrectCaloClusters.cpp
@@ -87,9 +87,10 @@ StatusCode CorrectCaloClusters::initialize() {
     return StatusCode::FAILURE;
   }
 
-  // Prepare upstream and downstream correction functions
+  // Prepare upstream, downstream and benchmark method correction functions
   initializeCorrFunctions(m_upstreamFunctions, m_upstreamFormulas, m_upstreamParams, "upstream");
   initializeCorrFunctions(m_downstreamFunctions, m_downstreamFormulas, m_downstreamParams, "downstream");
+  initializeBenchmarkCorrFunctions(m_benchmarkFunctions, m_benchmarkFormulas, m_benchmarkParams, "benchmark");
 
   info() << "Initialized following upstream correction functions:" << endmsg;
   for (size_t i = 0; i < m_upstreamFunctions.size(); ++i) {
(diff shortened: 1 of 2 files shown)


BrieucF:
Thanks Michaela! To be followed up:
- [ ] Add this algorithm to a CI config
- [ ] Improve consistency between (and be more explicit about) the way the correction formula is derived in `CalibrateBenchmarkMethod` and applied here (e.g. if derived on up/downstream corrected cluster the already corrected clusters should be used here)

RecCalorimeter/src/components/CorrectCaloClusters.h
```diff
@@ -177,6 +206,24 @@ class CorrectCaloClusters : public GaudiAlgorithm {
...
+
+  /// Benchmark method correction parameters (a, b, c, ...)
+  Gaudi::Property<std::vector<double>> m_benchmarkParams {
+      this, "benchmarkParameters", {}, "Benchmark parameters"};
+  /// Downstream correction formulas in ROOT notation
+  Gaudi::Property<std::vector<std::string>> m_benchmarkFormulas {
+      this, "benchmarkFormulas", {}, "Benchmark formulas (in ROOT notation)"};
+
+  /// benchmarkParamsApprox are used for the first estimate of the energy, below parameters for 100 GeV pion
+  Gaudi::Property<std::vector<double>> m_benchmarkParamsApprox {
```
kjvbrt, line 218: Are those initial values to the correction?

RecCalorimeter/src/components/CorrectCaloClusters.h
```diff
@@ -81,6 +85,16 @@ class CorrectCaloClusters : public GaudiAlgorithm {
                                      std::vector<std::vector<double>> parameters,
                                      const std::string& funcNameStem = "upDown");
 
+  /**
+   * Initialize vectors of benchmark correction functions.
+   *
+   * @return                  Status code.
+   */
```
kjvbrt, line 92: Can this function have the same name? it operates on vector of functions instead of vector of vector.

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -135,21 +145,29 @@ StatusCode CorrectCaloClusters::execute() {
   }
 
   // Apply upstream correction
-  {
+  if (m_upstreamCorr){
```
kjvbrt, line 148: Can also just check if the vector of functions is empty.

PR #39, opened 2023-03-16
Add cell filtering functionalities

BrieucF:
- Add the possibility to filter cells without adding noise
- Add the possibility to filter cells based on energy absolute value


BrieucF:
Now that the cell filtering functionality is not fully bounded to the noise application anymore, one could think of factorizing it as a separated tool that would be called by `CreateCaloCells` in a later PR.

PR #41, opened 2023-05-25
Use new MetaDataHandles to be compliant with Frame based I/O

tmadlener:
https://github.com/key4hep/k4FWCore/pull/100 makes the default DataSvc use the Frame, and some of the interfaces for accessing meta data (e.g. cellID strings) will change. This PR changes the access / writing of the cell ID encoding strings to use the new `MetaDataHandle` interface.

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CreateCaloCellPositionsFCCee.h
+++ b/RecFCCeeCalorimeter/src/components/CreateCaloCellPositionsFCCee.h
@@ -3,6 +3,7 @@
 
 // FCCSW
 #include "k4FWCore/DataHandle.h"
+#include "k4FWCore/MetaDataHandle.h"
 #include "k4Interface/ICellPositionsTool.h"
 
 // Gaudi
@@ -91,10 +92,10 @@ private:
   dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder("system:4");
   /// Input collection
   DataHandle<edm4hep::CalorimeterHitCollection> m_hits{"hits/hits", Gaudi::DataHandle::Reader, this};
+  MetaDataHandle<std::string> m_hitsCellIDEncoding{m_hits, "CellIDEncodingString", Gaudi::DataHandle::Reader, this};
   /// Output collection
   DataHandle<edm4hep::CalorimeterHitCollection> m_positionedHits{"hits/positionedHits", Gaudi::DataHandle::Writer, this};
-  PodioDataSvc* m_podioDataSvc;
-  ServiceHandle<IDataProviderSvc> m_eventDataSvc;
+  MetaDataHandle<std::string> m_positionedHitsCellIDEncoding{m_positionedHits, "CellIDEncodingString", Gaudi::DataHandle::Reader, this};
 
   int m_system_id = m_decoder->index("system");
   std::unordered_map<dd4hep::DDSegmentation::CellID, edm4hep::Vector3f> m_positions_cache{};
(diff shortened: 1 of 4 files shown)


kjvbrt:
The MetaDataHandle does not take reference to the `fatherAlg`.

RecFCCeeCalorimeter/src/components/CreateCaloCellPositionsFCCee.h
```diff
@@ -91,10 +92,10 @@ class CreateCaloCellPositionsFCCee : public GaudiAlgorithm {
   dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder("system:4");
   /// Input collection
   DataHandle<edm4hep::CalorimeterHitCollection> m_hits{"hits/hits", Gaudi::DataHandle::Reader, this};
+  MetaDataHandle<std::string> m_hitsCellIDEncoding{m_hits, "CellIDEncodingString", Gaudi::DataHandle::Reader, this};
```
kjvbrt, line 95: Should MetaDataHandle take `fatherAlg`?

RecFCCeeCalorimeter/src/components/CreateCaloCellPositionsFCCee.h
```diff
@@ -91,10 +92,10 @@ class CreateCaloCellPositionsFCCee : public GaudiAlgorithm {
...
   /// Output collection
   DataHandle<edm4hep::CalorimeterHitCollection> m_positionedHits{"hits/positionedHits", Gaudi::DataHandle::Writer, this};
-  PodioDataSvc* m_podioDataSvc;
-  ServiceHandle<IDataProviderSvc> m_eventDataSvc;
+  MetaDataHandle<std::string> m_positionedHitsCellIDEncoding{m_positionedHits, "CellIDEncodingString", Gaudi::DataHandle::Reader, this};
```
kjvbrt, line 98: Also here.

RecFCCeeCalorimeter/src/components/CreateCaloCellPositionsFCCee.h
tmadlener: Ah no. It seems like I didn't update this PR after that has been removed again from key4hep/k4FWCore#100. Let me fix that

kjvbrt:
The PR build locally with the new frame I/O

tmadlener:
This probably broke here: https://github.com/HEP-FCC/k4RecCalorimeter/pull/43/files

kjvbrt:
The #43 was intended to fix the issue. The `m_positionedHitsCellIDEncoding` should be output metadata handle no?

tmadlener:
Ah right. My bad then. Are there some simple examples that could be put together into a minimal test suite?

tmadlener:
Ah right. My bad then. Are there some simple examples that could be put together into a minimal test suite?

tmadlener:
The warnings can be ignored for now. This is effectively a small issue in the lookup logic in podio. However, since there is no schema evolution to do at the moment, everything works, even with these warnings. We can silence them if we have sorted things out if necessary.

giovannimarchiori:
> The segmentation violation issue should be fixed by [HEP-FCC/k4SimGeant4#40](https://github.com/HEP-FCC/k4SimGeant4/pull/40).

Indeed, I merged the changes of that MR in my local fork and I have been able to run the simulation successfully in a recent nightly and produce the same hits and cells as with older nightlies and the legacy podio interface

PR #48, opened 2023-08-28
Implement module-theta readout (with per-layer merging)

giovannimarchiori:
This code implements the cell positioning calculations for the module-theta merged readout implemented in https://github.com/HEP-FCC/FCCDetectors/pull/56

Tagging @gartrog , @BrieucF

### DIFF
--- a/RecCalorimeter/CMakeLists.txt
+++ b/RecCalorimeter/CMakeLists.txt
@@ -28,9 +28,6 @@ install(TARGETS k4RecCalorimeterPlugins
 
 include(CTest)
 
-add_test(NAME TemporaryTest
-         COMMAND k4run -h)
-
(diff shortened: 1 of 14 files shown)


kjvbrt:
Hi @giovannimarchiori,
since the new way how to calculate cell positions guts the original tool, would it be possible to instead create a new one? The interface for the cell positioning tools is defined in [ICellPositionsTool.h](https://github.com/key4hep/k4FWCore/tree/main/k4Interface/include/k4Interface/ICellPositionsTool.h).

giovannimarchiori:
Hi @kjvbrt ,
could please elaborate more? I don't understand exactly what's the problem here. I took a code that was working only for a specific readout (eta-phi) and made it more general (not it works for eta-phi but also for theta-phi and for the theta-module merged readout). Are you saying it's better instead to have separate cell position tools for different readout so that each tool is easier to understand ? In that case, I can create a new tool CellPositionsECalBarrelFCCeeTool (or something like that) in RecFCCeeCalorimeter

kjvbrt:
Yes, to have different tools for different readouts, this way the change is visible in the steering file.

giovannimarchiori:
Hi @kjvbrt ,
I created a new tool as you suggested

PR #51, opened 2023-09-27
make topoclustering work with new module-theta merged readout

giovannimarchiori:
This PR and related ones for FCCDetectors and LAr_scripts make topoclustering work with the new theta-module readout for the FCC-ee calorimeter.
I would like to initiate the review to discuss the code structure - I will then remove commented code and debug messages later. 
One thing that will probably be debated: I modified the ReadNoiseFromFileTool work for both the phi-eta and module-theta readouts work without creating a new class, but simply adding the possibility to pass to the tool a cell positioning tool, and read from there the eta or theta of cells.
If we decide to go this way, I will polish the code and add a property to tell the tool if it should assume eta or theta bins (in the current quick and dirty implementation, that is backward compatible, it is assumed that if segmentation is used then noise is histogrammed vs |eta|, otherwise vs theta). Otherwise I can create a separate class to read histograms vs theta, but with 95% the same code as that of ReadNoiseFromFileTool ..

Tagging @BrieucF @gartrog @dasphy

Related PRs:
https://github.com/HEP-FCC/FCCDetectors/pull/63
https://github.com/BrieucF/LAr_scripts/pull/13

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
+++ b/RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
@@ -1,0 +1,489 @@
+#include "CreateFCCeeCaloNeighbours.h"
+
+#include "DD4hep/Detector.h"
+#include "DetCommon/DetUtils.h"
+#include "k4Interface/IGeoSvc.h"
+
+#include "TFile.h"
+#include "TTree.h"
+
+DECLARE_COMPONENT(CreateFCCeeCaloNeighbours)
+
+CreateFCCeeCaloNeighbours::CreateFCCeeCaloNeighbours(const std::string& aName, ISvcLocator* aSL)
+    : base_class(aName, aSL) {
+  declareProperty( "outputFileName", m_outputFileName, "Name of the output file");
+}
+
+CreateFCCeeCaloNeighbours::~CreateFCCeeCaloNeighbours() {}
+
+StatusCode CreateFCCeeCaloNeighbours::initialize() {
+  // Initialize necessary Gaudi components
+  if (Service::initialize().isFailure()) {
+    error() << "Unable to initialize Service()" << endmsg;
+    return StatusCode::FAILURE;
+  }
+  m_geoSvc = service("GeoSvc");
+  if (!m_geoSvc) {
+    error() << "Unable to locate Geometry Service. "
+            << "Make sure you have GeoSvc and SimSvc in the right order in the configuration." << endmsg;
+    return StatusCode::FAILURE;
+  }
+  std::unordered_map<uint64_t, std::vector<uint64_t>> map;
+ 
+  // will be used for volume connecting
+  int eCalLastLayer;
+  std::pair<int, int> extremaECalLastLayerModule;
+  std::pair<int, int> extremaECalLastLayerTheta;
+  double eCalThetaOffset = 0;
+  double eCalThetaSize = 0;
+  //double eCalPhiOffset = 0;
+  double eCalModuleSize = 0;
+  double hCalThetaOffset = 0;
+  double hCalThetaSize = 0;
+  double hCalPhiOffset = 0;
+  dd4hep::DDSegmentation::BitFieldCoder* decoderECalBarrel = nullptr;
+  // will be used for volume connecting
+  std::pair<int, int> extremaHCalFirstLayerPhi;
+  std::pair<int, int> extremaHCalFirstLayerTheta;
+  std::pair<int, int> extremaHCalFirstLayerZ;
+  dd4hep::DDSegmentation::BitFieldCoder* decoderHCalBarrel = nullptr;
+
+  ////////////////////////////////////
+  /// SEGMENTED THETA-MODULE VOLUMES  ///
+  ////////////////////////////////////
+  
+  for (uint iSys = 0; iSys < m_readoutNamesSegmented.size(); iSys++) {
+    // Check if readouts exist
+    info() << "Readout: " << m_readoutNamesSegmented[iSys] << endmsg;
+    if (m_geoSvc->getDetector()->readouts().find(m_readoutNamesSegmented[iSys]) == m_geoSvc->getDetector()->readouts().end()) {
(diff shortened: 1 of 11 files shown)


BrieucF:
Hi Giovanni, thanks! 

If I understand correctly what you want to discuss, I see three possibilities:
1) We go for the if's as you did, knowing that if we try a cos(theta) segmentation or yet something else we will have a list of growing if's, probably not nice
2) We create a new tool under RecFCCeeCalorimeter. The argument in favor of this is that the theta segmentation does not make sense for FCC-hh and will never be used in this context. It also minimize the modifications we apply to the code that was used for FCC-hh studies (maximizing thus the chances that it will still work for FCC-hh).
3) We find a way, with an additional layer of abstraction, to have the same treatment regardless of the eta/theta segmentation. This is probably the nicest solution (avoiding if and code duplication) but I don't know if it is at all possible.

kjvbrt:
Hi all,
I would argue for separate tools/algorithms under `RecFCCeeCalorimeter` for theta segmentation, since it is not usable for FCChh, as already mentioned. But also, in my opinion, the tools/algorithms should be disposable.

giovannimarchiori:
Hi @BrieucF and @kjvbrt ,
I just pushed some new code, to restore the original ReadNoiseFromFileTool and move my corresponding code for the theta-based readout to RecFCCeeCalorimeter.
Do you think the organisation of the code in this PR (abstracting for a moment for the low-level technical implementation) is OK or do I need to refactor the code in a different way?
Also, what about https://github.com/HEP-FCC/FCCDetectors/pull/63/, is it OK to add those extra functions to DetCommon/DetUtils?
if yes then I will then focus on completing the missing things (treatment of diagonal neighbours + cleanup of comments and debug statements)
Thanks!

kjvbrt:
Hi @giovannimarchiori, I like the organisation of the code.

Do I understand correctly that these algorithms will be used in three different workflows?
First, to create map of neighbours. Second to derive noise and the last one to use both map of neighbours and noise in the reconstruction?
Can you provide example steering files for each?

This is a nitpick, but might be good to solve now. @BrieucF what naming convention for the algorithms would be most compatible with k4Geo? I mean, where to put `FCCee` if at all.

giovannimarchiori:
HI @kjvbrt 

one has to proceed as described here https://github.com/giovannimarchiori/LAr_scripts/tree/main/geometry but using the _theta scripts in this PR (https://github.com/BrieucF/LAr_scripts/pull/13) instead:
# create noise map
1) python create_capacitance_file_theta.py
2) python create_noise_file_chargePreAmp_theta.py
3) fccrun noise_map_theta.py
# create neighbours map
4) fccrun neighbours_theta.py
# run the reconstruction
5) fccrun run_thetamodulemerged.py

BrieucF:
> This is a nitpick, but might be good to solve now. @BrieucF what naming convention for the algorithms would be most compatible with k4Geo? I mean, where to put `FCCee` if at all.

Hi @kjvbrt, "if at all" will depend whether or not other calorimeters (non-FCC) use this repository. Since for now the answer is no I don't think we should try to tackle this in this PR.

giovannimarchiori:
Hi @kjvbrt  I think this PR and the related ones are now good for review. I have added the treatment of the diagonal neighbours, cleaned up the code, removed debugging statements, updated the detector parameters to get back 1536 modules. main thing missing is barrel ECAL-HCAL matching (to find neighbours across different volumes) but this can be tackled by a separate PR once we will start looking at hadronic clusters.

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -0,0 +1,488 @@
...
+        }
+      }
+    }
+  }
+
+  //////////////////////////////////////////////////
+  ///      BARREL: connection ECAL + HCAL        ///
+  /////////////////////////////////////////////////
+  int count=0;
+/*
```
kjvbrt, line 300: Is this peace of code intended to be used for ECAL-HCAL matching?

giovannimarchiori: this part needs to be implemented. I copied here the code that was pre-existing but I am gonna refactor it in the coming days, as soon as I find out the settings to use for the HCAL clustering.

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNoiseLevelMap.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return StatusCode::FAILURE;
+  }
+  m_geoSvc = service("GeoSvc");
+  if (!m_geoSvc) {
+    error() << "Unable to locate Geometry Service. "
+            << "Make sure you have GeoSvc and SimSvc in the right order in the configuration." << endmsg;
+    return StatusCode::FAILURE;
+  }
+
+  //m_ecalBarrelNoiseTool.retrieve();
```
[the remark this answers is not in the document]
giovannimarchiori: This class is inspired by the one that was pre-existing here: https://github.com/HEP-FCC/k4RecCalorimeter/blob/a848093fc7008c1466656afe49170bbf5e552d60/RecFCChhCalorimeter/src/components/CreateFCChhCaloNoiseLevelMap.cpp
In that class there is no retrieve of the two noise tools and I was surprised, but I wasn't sure that this should be added, so in my class I put these two commented lines with the retrieve of the two tools. Do you think I need to explicitly retrieve the tools and check that it was successful?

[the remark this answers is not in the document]
giovannimarchiori: added check in latest commit

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
kjvbrt: OK

kjvbrt:
The failure of the CI tests expected

PR #68, opened 2024-02-27
Add Gaudi algorithm to perform MVA calibration of cluster energies

giovannimarchiori:
Implement a Gaudi algorithm that reads a model via ONNX runtime and calibrates the clusters
Pre-trained models for ALLEGRO's ECAL will be uploaded in the LAr_scripts package (and FCC-config)


RecFCCeeCalorimeter/CMakeLists.txt
```diff
@@ -2,6 +2,13 @@
 # Package: RecFCCeeCalorimeter
 ################################################################################
 
+# ONNX DEPENDENCIES
```
kjvbrt, line 5: can you put `find_package` to the top CMake file? Also update list of dependencies in the README.

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  // Check if readouts exist
+  {
+    bool readoutMissing = false;
+    for (size_t i = 0; i < m_readoutNames.size(); ++i)
+    {
+      auto readouts = m_geoSvc->getDetector()->readouts();
+      if (readouts.find(m_readoutNames.value().at(i)) == readouts.end())
+      {
+        readoutMissing = true;
```
kjvbrt, line 59: You can directly return `StatusCode::FAILURE` and not use `readoutMissing` variable.

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -0,0 +1,368 @@
...
+  for (auto &s : input_shapes)
+  {
+    if (s < 0)
+    {
+      s = 1;
+    }
+  }
+
+  // print name/shape of outputs
+  // std::vector<std::string> output_names;
```
kjvbrt, line 231: Delete?

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.h
```diff
@@ -0,0 +1,142 @@
...
+  //    this, "calibrationFiles", {}, "Files with the calibration parameters"};
+  Gaudi::Property<std::string> m_calibrationFile {
+      this, "calibrationFile", {}, "File with the calibration parameters"};
+
+  // total number of layers summed over the various subsystems
+  // should be equal to the number of input features of the MVA
+  size_t m_numLayersTotal;
+
+  // the ONNX runtime session for applying the calibration,
+  // the environment, and the input and output shapes and names
```
kjvbrt, line 133: Can those member variables have `m_` in front of their names?

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.h
kjvbrt, line 1: Can you try to inherit from new `Gaudi::Algorithm` instead of `GaudiAlgorithm`? `Gaudi::Algorithm` lives in  header: `GaudiKernel/Algorithm.h`.
Caution: The datahandle need to be declared mutable.
Example of the algorthm using new style of algorithms: https://github.com/kjvbrt/k4Gen/blob/main/k4Gen/src/components/GenEventFilter.h

giovannimarchiori: compilation fails because I have to declare the execute(..) method as const, which is incompatible with the line
  edm4hep::ClusterCollection *outClusters = initializeOutputClusters(inClusters);

what should I do?

RecFCCeeCalorimeter/CMakeLists.txt
giovannimarchiori: fixed in latest commit

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
giovannimarchiori: fixed in latest commit

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.h
giovannimarchiori: fixed in latest commit

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.h
giovannimarchiori: done in latest commit

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
giovannimarchiori: fixed in latest commit

giovannimarchiori:
Hi @kjvbrt , 
thanks for your careful review!
I implemented all your suggestions in the latest commit
cheers,
  Giovanni

BrieucF:
This is really great to have! Thanks Giovanni!

For the trained model, I propose to centralize them here `/eos/project/f/fccsw-web/www/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v02/`, following the same versioning/path conventions as in `k4geo`

giovannimarchiori:
Hi @BrieucF , wouldn't it be better to put the files in GitHub, so that we use its versioning system, record changes in the commit messages, can track more easily when files were updated, and so on? They are small files, they are not going to make the repository too big

BrieucF:
If they are small enough why not

giovannimarchiori:
Those from LGBM, which has a performance very similar to XGB, are 300-400k. Those from XGB are about 2.5M

BrieucF:
On the other hand the diff of those files will always be a "full diffs" so not sure the gitHub versioning really brings something more than having our models as `model_v1`, `model_v2`, etc. It will pollute the gitHub history and if we introduce bigger models at some point we will end up with heterogenous storage places. NB: the files from `/eos/project/f/fccsw-web/www/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v02/` can be retrieved with `wget` or `curl`, e.g.:

```
curl -O -L http://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v02/elecNoise_ecalBarrelFCCee_theta.root
```

PR #70, opened 2024-03-21
Removing m_allCells from CaloTopoClusterFCCee algorithm members

kjvbrt:
The `m_allCells` unordered map is to my mind unnecessary as it does not provide any boost in performance, and introduces compulsory zeroing of the map before every event.
Little benchmark on 100 evt:
```
Runtime:
before [s]: 8.456, 8.603, 8.489, 8.452, 8.661
after [s]: 8.492, 8.578, 8.548, 8.398, 8.568
average before[s]: 8.5322
average after[s]: 8.5168
```

Also, fixing issue/crash when running over `ddsim` generated sample(s).

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
+++ b/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
@@ -1,30 +1,26 @@
 #include "CaloTopoClusterFCCee.h"
 #include "../../../RecCalorimeter/src/components/NoiseCaloCellsFromFileTool.h"
 
+// std
+#include <algorithm>
+#include <memory>
+#include <numeric>
+#include <unordered_set>
+#include <vector>
+
 // k4geo
 #include "detectorCommon/DetUtils_k4geo.h"
 
 // k4FWCore
 #include "k4Interface/IGeoSvc.h"
 
-// edm4hep
-#include "edm4hep/Cluster.h"
+// EDM4hep
 #include "edm4hep/ClusterCollection.h"
-#include "edm4hep/CalorimeterHit.h"
 #include "edm4hep/CalorimeterHitCollection.h"
 
 // DD4hep
 #include "DD4hep/Detector.h"
 #include "DD4hep/Readout.h"
-
-// std C++ includes
-#include <algorithm>
-#include <map>
-#include <memory>
-#include <numeric>
-#include <unordered_map>
-#include <unordered_set>
-#include <vector>
 
 DECLARE_COMPONENT(CaloTopoClusterFCCee)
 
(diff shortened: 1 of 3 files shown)


BrieucF:
Thanks Juraj, tagging also @giovannimarchiori  and @gartrog

giovannimarchiori:
Hi @kjvbrt ,
since you're touching this code, could you please have a look at why the positions of topoclustered cells in the output root file are in cm while those of other hits/cells/cells after SW clustering are in mm?
I suspect it's this line in CaloTopoClusterFCCee.cpp
     newCell.setPosition(edm4hep::Vector3f(posCell.X(), posCell.Y(), posCell.Z()));
where one should divide the 3 coordinates by edm4hep::mm (which is = 0.1)?

Thanks a lot!
  Giovanni

kjvbrt:
Hi @giovannimarchiori,
good point, the cell position comes from DD4hep function, which uses cm as base unit. Using dd4hep::mm ...

What about the position of the cluster itself? Is it OK to have it in cm or should I change that as well?

Best,
Juraj

giovannimarchiori:
right, also the cluster position is in different units for topoclusters (cm) and sw clusters (mm) so would probably a good idea to harmonise and divide cluster position by /mm here

RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
```diff
@@ -200,15 +200,16 @@ StatusCode CaloTopoClusterFCCee::execute() {
...
-      
+      edmClusterCells->push_back(newCell);
+      auto er = allCells.erase(cID);
+
       if (er!=1)
               info() << "Problem in erasing cell ID from map." << endmsg;
     }
     cluster.setEnergy(energy);
-    cluster.setPosition(edm4hep::Vector3f(posX / energy, posY / energy, posZ / energy));
+    cluster.setPosition(edm4hep::Vector3f{posX / energy, posY / energy, posZ / energy});
```
giovannimarchiori, line 212: please divide all 3 coordinates by  dd4hep::mm

giovannimarchiori, line 212:
```suggestion
    cluster.setPosition(edm4hep::Vector3f{posX / energy / dd4hep::mm, 
                                          posY / energy / dd4hep::mm,
                                          posZ / energy / dd4hep::mm});
```

giovannimarchiori:
Hello! May I ask what's the status of the review of this PR?

Thanks, Giovanni

BrieucF:
After homogenization of the units it looks good to me!

PR #72, opened 2024-03-21
AugmentClustersFCCee - Clusters barycenters layered correction

AlexisMlz:
Added script elements `AugmentClustersFCCee` to apply logarithmic layered correction to ECal clusters barycenters calculation and save information (E, Theta, Phi) as shape parameters.

No change has been made to the existing part of the package.


giovannimarchiori:
Thanks @AlexisMlz !

Tagging @BrieucF @gartrog 

Additional info: the log(E) weights are only applied for theta, and if negative offsets are provided, E weights are used instead. For phi, always E-weights are used

@AlexisMlz, please submit also the PR for the fccrun steering file for https://github.com/BrieucF/LAr_scripts/tree/main .

giovannimarchiori:
Hello! May I ask what's the status of the review of this PR? Once it's in we could develop something like this (on top of it or in a separate alg) to compute lateral shower shapes and save them as shapeParameters of the clusters

Thanks, Giovanni

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,255 @@
...
+  if (showerShapeDecorations != "")
+    showerShapeDecorations += ":";
+  for (size_t k = 0; k < m_detectorNames.size(); k++)
+  {
+    const char *detector = m_detectorNames[k].c_str();
+    for (unsigned layer = 0; layer < m_numLayers[k]; layer++)
+    {
+      if (layer > 0)
+        showerShapeDecorations += ":";
+      showerShapeDecorations += Form("f%s%d:theta%s%d:phi%s%d",
```
kjvbrt, line 81: Can you avoid using free floating `Form()`?

kjvbrt:
Hi @giovannimarchiori, @AlexisMlz, do you have an example how the algorithm is used in the Gaudi steering file somewhere?

giovannimarchiori:
Yes, here: https://github.com/BrieucF/LAr_scripts/pull/23/files

kjvbrt:
The files are in DOS format, can you convert them with `dos2unix`?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+StatusCode AugmentClustersFCCee::finalize()
+{
+  return Gaudi::Algorithm::finalize();
+}
+
+StatusCode AugmentClustersFCCee::execute(const EventContext &evtCtx) const
+{
+  (void)evtCtx; // event context not used
```
kjvbrt, line 97: This is here just to avoid warning? Can you instead comment out the variable name?
```
StatusCode AugmentClustersFCCee::execute(const EventContext & /* evtCtx */) const
```

kjvbrt: Or probably better: `[[maybe_unused]]` (https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)

AlexisMlz:
Thanks for the comments, I should have fixed the files accordingly. I'll also convert the steering file that probably is also in dos format..

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,253 @@
...
+  if (showerShapeDecorations != "")
+    showerShapeDecorations += ":";
+  for (size_t k = 0; k < m_detectorNames.size(); k++)
+  {
+    const char *detector = m_detectorNames[k].c_str();
+    for (unsigned layer = 0; layer < m_numLayers[k]; layer++)
+    {
+      if (layer > 0)
+        showerShapeDecorations += ":";
+      showerShapeDecorations += TString::Form("f%s%d:theta%s%d:phi%s%d",
```
kjvbrt, line 81: `TString::Format()` should work :)

giovannimarchiori: Hi @kjvbrt I haven't tested this change myself, but the original Form(..) method in global namespace was returning a  const char*, which we can then add to showerShapeDecorations (which is a std::string).
Not sure that with TString::Form or TString::Format (that both return a TString) the code will still compile. @AlexisMlz did you try to compile the code?

See https://root.cern.ch/doc/master/classTString.html#aeb44d6183d8b1f1b7090dbd3c93f5e39

Note: this is not to be confused with [Format](https://root.cern.ch/doc/master/classTString.html#ac34bbb89627e20624a416dfbb7a3db70) and [Form](https://root.cern.ch/doc/master/classTString.html#aeb44d6183d8b1f1b7090dbd3c93f5e39) (in the global namespace) which returns a const char* and relies on a thread-local static character buffer.

BrieucF:
Can we add this algorithm to the integration test `RecFCCeeCalorimeter/tests/options/runCaloSim.py`?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
giovannimarchiori: BTW what was the problem with using Form(..) in first place ? 
https://root.cern.ch/doc/master/TString_8cxx_source.html#l02489

kjvbrt:
> Can we add this algorithm to the integration test `RecFCCeeCalorimeter/tests/options/runCaloSim.py`?

This would be much welcomed :)

giovannimarchiori:
I see that runCaloSim.py uses ALLEGRO o1_v01, while this code was developed with v02 in mind...  though we tried to make it as much segmentation-agnostic as possible

BrieucF:
> I see that runCaloSim.py uses ALLEGRO o1_v01, while this code was developed with v02 in mind... though we tried to make it as much segmentation-agnostic as possible

Yes, we can update it to o1_v02 if needed, or create another minimalist config file producing the needed object to run `AugmentCluster`. The goal is just to make sure that the algorithm actually runs (in addition to the test already checking that it compiles).

giovannimarchiori:
We tested that it runs with the script in the other PR: 
https://github.com/BrieucF/LAr_scripts/pull/23/files

BrieucF:
Yes please, the idea is to make sure nothing will break it in the future. The fact that it runs today, I trust you of course ;-)

giovannimarchiori:
Test is ready for v03, we can put this in when the k4geo PR for ALLEGRO v03 is merged

BrieucF:
Regarding the metadata with the description of the shapeParameters, can you try to have a vector of strings instead of a single string with semicolon separation?

In the header: 
```
MetaDataHandle<std::vector<std::string>>
```
And in the .cpp
```
std::vector<std::string> showerShapeDecorations;
```

Then adapt the filling of `showerShapeDecorations`.

giovannimarchiori:
OK so the latest push by @AlexisMlz includes the metadata as vector of strings, plus the test for this new alg (using ECal v03)
@kjvbrt , @BrieucF , do you understand why the build tests fail? The error message is "In file included from /Package/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp:1:
/Package/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h:53:14: error: 'shapeParameterNames' is not a member of 'edm4hep'
   53 |     edm4hep::shapeParameterNames,
      |              ^~~~~~~~~~~~~~~~~~~"

I used to get this when using an old nightly that didn't have an updated version of edm4hep that includes shapeParameterNames (https://github.com/key4hep/EDM4hep/blob/792cefc3a0f7cb3dfa12aba50a24a6f37089e84e/include/edm4hep/Constants.h#L26) but since at least the day when we opened the PR everything compiles fine locally

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
```diff
@@ -0,0 +1,85 @@
...
+{
+  namespace DDSegmentation
+  {
+    class BitFieldCoder;
+  }
+}
+
+/** @class AugmentClustersFCCee
+ *
+ *  Add to the cluster shape parameters the energies and barycenter positions
```
BrieucF, line 29:
```suggestion
 *  Add to the cluster shape parameters the sum of the cluster cells energy and barycenter theta/phi coordinates
```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // initialise the list of metadata for the clusters
+  // append to the metadata of the input clusters
+  std::vector<std::string> showerShapeDecorations = m_inShapeParameterHandle.get();
+  for (size_t k = 0; k < m_detectorNames.size(); k++)
+  {
+    const char *detector = m_detectorNames[k].c_str();
+    for (unsigned layer = 0; layer < m_numLayers[k]; layer++)
+    {
+      showerShapeDecorations.push_back(Form("f%s%d", detector, layer));
```
BrieucF, line 77:
```suggestion
      showerShapeDecorations.push_back(Form("energy_%s_layer_%d", detector, layer));
```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,252 @@
...
+      showerShapeDecorations.push_back(Form("theta%s%d", detector, layer));
```
BrieucF, line 78:
```suggestion
      showerShapeDecorations.push_back(Form("theta_%s_layer_%d", detector, layer));
```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -0,0 +1,252 @@
...
+      showerShapeDecorations.push_back(Form("phi%s%d", detector, layer));
```
BrieucF, line 79:
```suggestion
      showerShapeDecorations.push_back(Form("phi_%s_layer_%d", detector, layer));
```

RecFCCeeCalorimeter/tests/options/run_thetamodulemerged.py
```diff
@@ -0,0 +1,779 @@
...
+createTopoInput.ecalEndcapCells.Path = "emptyCaloCells"
+createTopoInput.ecalFwdCells.Path = "emptyCaloCells"
+createTopoInput.hcalBarrelCells.Path = hcalBarrelPositionedCellsName2
+createTopoInput.hcalExtBarrelCells.Path = "emptyCaloCells"
+createTopoInput.hcalEndcapCells.Path = "emptyCaloCells"
+createTopoInput.hcalFwdCells.Path = "emptyCaloCells"
+cellPositionHcalBarrelNoSegTool = None
+cellPositionHcalExtBarrelTool = None
+
+neighboursMap = "/LAr_scripts/data/neighbours_map_v03_ecalB_thetamodulemerged.root"
```
BrieucF, line 594: This will likely not work (the root file will not be found), or am I missing something? Can you maybe streamline this config file to create simple clusters (e.g. with sliding window without the need for these rootfiles) so that the test run?

BrieucF (commented):
Merci Alexis!

BrieucF:
> I implemented it locally and the metadata seems to be filled correctly:
> 
> `AugmentedCaloClusters__shapeParameterNames[fEMB0, thetaEMB0, phiEMB0, fEMB1, thetaEMB1, phiEMB1, fEMB2, thetaEMB2, phiEMB2, fEMB3, thetaEMB3, phiEMB3, fEMB4, thetaEMB4, phiEMB4, fEMB5, thetaEMB5, phiEMB5, fEMB6, thetaEMB6, phiEMB6, fEMB7, thetaEMB7, phiEMB7, fEMB8, thetaEMB8, phiEMB8, fEMB9, thetaEMB9, phiEMB9, fEMB10, thetaEMB10, phiEMB10]`
> 
> @AlexisMlz can you please update this PR with the files AugmentClustersFCCee.cpp AugmentClustersFCCee.h CMakeLists.txt run_thetamodulemerged.py in my public/ directory on our APC server, so that we move on with the merge of this PR?
> 
> BTW I had to move back from TString::Format fo Form(..) since the code would not compile (no explicit cast from TString to std::string.
> 
> Thanks, Giovanni

Thanks. You were probably looking at the test in the stable release, they are expected to fail (not for long though because we have a new release). The test with nightlies fail because of FCCBASEDIR aqnd the rootfile that are needed for e.g. the noise: https://github.com/HEP-FCC/k4RecCalorimeter/actions/runs/8647156803/job/23707991037?pr=72#step:3:384 . See my comment about streamlining the test steering file

RecFCCeeCalorimeter/tests/options/run_thetamodulemerged.py
giovannimarchiori: this actually works because in the following lines the topoclustering algorithms are not scheduled, see eg https://github.com/HEP-FCC/k4RecCalorimeter/pull/72/files#diff-262cb6467fba07c8f5498412934192dd3a79eec0225820307b2b8464a1538b07R748

We just took our full script and commented away a few lines to make the test run, rather than spending more time and drop everything that is not needed ;)

kjvbrt:
Hi @giovannimarchiori, @AlexisMlz
The current test provided will work only as part of the LAr_scripts. It depends on several root files from there and also on FCCBASEDIR environment variable defined there. The ROOT files can be hosted on the FCCSW web, but you need to shed the dependency on FCCBASEDIR.

giovannimarchiori:
Hi @kjvbrt, yes, you and @BrieucF are right, I had forgotten unsetting the FCCBASEDIR environment variable before running the test locally.

I have removed all topocluster tools that require FCCBASEDIR.

@AlexisMlz the new version of the file is in my public/ , can you push it to the branch for the PR?

Thanks, Giovanni

giovannimarchiori:
The test fails because ALLEGRO_o1_v03 is not found. It should work as soon as this PR is merged https://github.com/key4hep/k4geo/pull/332

BrieucF:
https://github.com/key4hep/k4geo/pull/332 was merged, we will though have to wait for tomorrow's nightlies

RecFCCeeCalorimeter/CMakeLists.txt
```diff
@@ -45,6 +45,12 @@ add_test(NAME FCCeeLAr_simulateForReco
 )
 set_test_env(FCCeeLAr_simulateForReco)
 
+add_test(NAME FCCeeLAr_simRecoAllegroV2
```
BrieucF, line 48: Shall we set the name to v3 instead of v2 since it actually tests `ALLEGRO_o1_v03` ?

giovannimarchiori: Changed by @AlexisMlz in latest commit

giovannimarchiori:
Tests run fine with nightly and test name was updated so it should be OK to merge ;) ?
And then with this comes the PR with the steering script https://github.com/BrieucF/LAr_scripts/pull/23 
Thanks!

BrieucF:
Thanks Alexis, thanks Giovanni!

PR #73, opened 2024-04-05
Cell position cast to float

kjvbrt:
In EDM4hep calohit position is float, but `dd4hep::Position` is double.


BrieucF:
Thanks Juraj, shall I merge this to have the test passing in https://github.com/HEP-FCC/k4RecCalorimeter/pull/72 ?

kjvbrt:
The tests in #72 do not fail because of this, but you can merge it :)

PR #74, opened 2024-04-15
update SW to run on HCal

mmlynari:
Added the possibility to run sliding window algorithm on HCal and ECal+HCal. 
Tagging @BrieucF , @giovannimarchiori and @dasphy


giovannimarchiori:
Very nice!

Question for you and @BrieucF : what is this Multi segmentation? is this a legacy from past code? Shall we keep it or get rid of it if it's not used anymore?

Thanks,
   Giovanni

BrieucF:
I would keep the multisegmentation check if it is not causing problems, so that is fine like this (in this PR it is done line 386)

PR #77, opened 2024-04-18
use shapeParameters in MVA calibration

giovannimarchiori:
This PR adds to the MVA calibration algorithm the possibility to read the energy fractions per layer directly from the shapeParameters of the input clusters (if available) rather than recalculating them on the fly. It also shows how to save in output the shapeParameters of the input collections plus some extra parameters (in this case the raw energy of the cluster before the MVA calibration).

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
+++ b/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
@@ -49,7 +49,7 @@ StatusCode CalibrateCaloClusters::initialize()
   }
 
   // Check if readouts exist
-  for (size_t i = 0; i < m_readoutNames.size(); ++i)
+  for (int i = 0; i < (int) m_readoutNames.size(); ++i)
   {
     auto readouts = m_geoSvc->getDetector()->readouts();
     if (readouts.find(m_readoutNames.value().at(i)) == readouts.end())
@@ -83,7 +83,7 @@ StatusCode CalibrateCaloClusters::initialize()
 
   // calculate total number of layers summed over the various subsystems
   m_numLayersTotal = 0;
-  for (size_t i = 0; i < m_numLayers.size(); ++i)
+  for (int i = 0; i < (int) m_numLayers.size(); ++i)
   {
     m_numLayersTotal += m_numLayers[i];
   }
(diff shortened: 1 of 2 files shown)


giovannimarchiori:
Usage example: very similar to before, just need to pass to the algorithm the string of the detector name used in the shapeParameters to label the energy fraction decorations, i.e. add the systemNames = ... row below:

```
calibrateCaloClusters = CalibrateCaloClusters("calibrateCaloClusters",
                                              inClusters="Augmented" + createClusters.clusters.Path,
                                              outClusters="Calibrated" + createClusters.clusters.Path,
                                              systemIDs=[4],
                                              systemNames=["EMB"],
                                              numLayers=[ecalBarrelLayers],
                                              firstLayerIDs=[0],
                                              readoutNames=[
                                                  ecalBarrelReadoutName],
                                              layerFieldNames=["layer"],
                                              calibrationFile="data/lgbm_calibration-CaloClusters-v03.onnx",
                                              OutputLevel=INFO
                                              )
```

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -49,7 +49,7 @@ StatusCode CalibrateCaloClusters::initialize()
   }
 
   // Check if readouts exist
-  for (size_t i = 0; i < m_readoutNames.size(); ++i)
+  for (int i = 0; i < (int) m_readoutNames.size(); ++i)
```
kjvbrt, line 52: Do you need `i` to be negative?

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -319,43 +351,70 @@ StatusCode CalibrateCaloClusters::calibrateClusters(const edm4hep::ClusterCollec
...
+        if (decoder->get(cellID, "system") != systemID)
+        {
+          continue;
+        }
+        int layer = decoder->get(cellID, layerField);
+        energiesInLayers[startPositionToFill + layer - firstLayer] += cell->getEnergy();
+      }
+    }
+    // divide by the cluster energy to prepare the inputs for the MVA
+    for (int k = 0; (int) k < m_numLayersTotal; ++k)
```
kjvbrt, line 413: Maybe `unsigned int`? as in many other places.

giovannimarchiori: changed to unsigned

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
giovannimarchiori: changed to unsigned

PR #78, opened 2024-04-22
Update benchmark calibration and correction

mmlynari:


RecCalorimeter/src/components/CalibrateBenchmarkMethod.cpp
```diff
@@ -150,126 +144,176 @@ StatusCode CalibrateBenchmarkMethod::execute() {
...
 }
 
 
 StatusCode CalibrateBenchmarkMethod::finalize() {
   // the actual minimisation is running here 
-  std::cout << "Running minimisation for " << m_vecEgenerated.size() << " #events! \n";
-  if (m_vecEgenerated.size() != m_vecEinECaltotal.size()){
+  std::cout << "Running minimisation for " << m_vecGeneratedEnergy.size() << " #events! \n";
+  if (m_vecGeneratedEnergy.size() != m_vecTotalEnergyinECal.size()){
     std::cout << "Something's wrong!!" << std::endl;
```
kjvbrt, line 242: Can you be more specific?

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -23,6 +25,7 @@
 
 // ROOT
 #include "TF2.h"
+#include <cmath>  // Include the <cmath> header for std::fabs
```
kjvbrt, line 28: Is cmath library part of ROOT?

RecCalorimeter/src/components/CorrectCaloClusters.h
```diff
@@ -177,6 +214,34 @@ class CorrectCaloClusters : public GaudiAlgorithm {
...
+      this, "benchmarkFormulas", {}, "Benchmark formulas (in ROOT notation)"};
+
+  /// Possibility to use different parametrization for very low energies (<10GeV) for benchmark method 
+  /// If true, the second formula in the input vector will be used for energies below m_eneSwitch 
+  Gaudi::Property<bool> m_benchmarkTwoFormulas{this, "benchmarkTwoFormulas", false};
+
+  /// For the energy (GeV) below this given threshold, the second energy formula in benchmarkFormulas is used 
+  /// To be used together with setting m_benchmarkTwoFormulas to true 
+  Gaudi::Property<double> m_benchmarkEneSwitch {
+      this, "benchmarkEneSwitch", {5.}, "Energy threshold in GeV to use different parametrization"};
```
kjvbrt, line 231: Can you merge this into one property? I mean is the second formula set won't be used, just set eneSwith below zero.

RecCalorimeter/src/components/CalibrateBenchmarkMethod.cpp
mmlynari: Thank you @kjvbrt  for your suggestions, I implemented all of them.

RecCalorimeter/src/components/CorrectCaloClusters.h
```diff
@@ -118,6 +138,19 @@ class CorrectCaloClusters : public GaudiAlgorithm {
                           int systemID,
                           int layerID);
 
+  /**
+   * Get sum of energy from cells in the whole calorimeter.
+   * This energy is not calibrated.
+   *
+   * @param[in]  cluster       Pointer to cluster of interest.
+   * @param[in]  readoutName   Name of the readout.
```
kjvbrt, line 146: systemID is also input parameter for this function.

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -318,6 +377,118 @@ StatusCode CorrectCaloClusters::applyDownstreamCorr(const edm4hep::ClusterCollec
...
+                                         m_readoutNames[ecal_index],
+                                         m_systemIDs[ecal_index]);
+
+    double totalEnergyInHCal = getTotalEnergy(inClusters->at(j),
+                                         m_readoutNames[hcal_index],
+                                         m_systemIDs[hcal_index]);
+                                         
+
+    // calculate approximate benchmark energy using non energy dependent benchmark parameters
+    double approximateBenchmarkEnergy = 0.0;
```
kjvbrt, line 424: zero in `approximateBenchmarkEnergy` is overwritten in next line. Can you remove this line?

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -318,6 +377,118 @@ StatusCode CorrectCaloClusters::applyDownstreamCorr(const edm4hep::ClusterCollec
...
+    double approximateBenchmarkEnergy = 0.0; 
+    approximateBenchmarkEnergy = m_benchmarkParamsApprox[0] * totalEnergyInECal +
+                                 m_benchmarkParamsApprox[1] * totalEnergyInHCal +
+                                 m_benchmarkParamsApprox[2] * sqrt(abs(energyInLastLayerECal * m_benchmarkParamsApprox[0] * energyInFirstLayerHCal * m_benchmarkParamsApprox[1])) +
+                                 m_benchmarkParamsApprox[3] * pow(totalEnergyInECal * m_benchmarkParamsApprox[0], 2) +
+                                 m_benchmarkParamsApprox[4] * energyInFirstLayerECal +
+                                 m_benchmarkParamsApprox[5];
+
+    // Calculate energy-dependent benchmark parameters p[0]-p[5]
+    int benchmarkFormulaIndexHighEne = 0;
```
kjvbrt, line 433: Can you instead of declaring an index create new vector out of only high energy formulas?
```
auto benchmarkFormulaHighEne = m_benchmarkFunctions.at(0);
```

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -318,6 +377,118 @@ StatusCode CorrectCaloClusters::applyDownstreamCorr(const edm4hep::ClusterCollec
...
+    int nParam = m_benchmarkFunctions.at(benchmarkFormulaIndexHighEne).size(); 
+    std::vector<double> benchmarkParameters(nParam,-1.); 
+
+    if (m_benchmarkEneSwitch>0.)
+    {
+      int benchmarkFormulaIndexLowEne = 1;
```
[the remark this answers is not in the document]
mmlynari: right, I added more safety checks in initialize()

RecCalorimeter/src/components/CorrectCaloClusters.cpp
mmlynari: done

RecCalorimeter/src/components/CorrectCaloClusters.h
mmlynari: added in the latest commit

RecCalorimeter/src/components/CorrectCaloClusters.cpp
mmlynari: it should be OK now with more safety checks

RecCalorimeter/src/components/CorrectCaloClusters.cpp
```diff
@@ -72,21 +77,66 @@ StatusCode CorrectCaloClusters::initialize() {
...
+        error() <<  "A single formula parametrization for benchmark correction is turned on, but the benchmarkFormulas[0] vector does not have the proper size, the size should be 6." << endmsg;
+        return StatusCode::FAILURE;
+      }
+    }
+    if (m_benchmarkEneSwitch>0.){
+      if (m_benchmarkFormulas.size()<2){
+        error() <<  "The two formulas parametrization for benchmark correction is turned on, but the benchmarkFormulas vector does not have the proper size, the size should be 2." << endmsg;
+        return StatusCode::FAILURE;
+      }
+      if (m_benchmarkFormulas[0].size() != m_benchmarkParamsApprox.size() || m_benchmarkFormulas[1].size() != m_benchmarkParamsApprox.size()){
```
kjvbrt, line 135: shouldn't this be `m_benchmarkFormulas[0]` or `m_benchmarkFormulas[1]`?

mmlynari: it could be, but then we would only check that the two vectors have the same size and we would not check that the size is correct. E.g. they both can still have the same size, e.g. 5, while we want 6. While in this case we first check that m_benchmarkParamsApprox has the correct size and use it as a reference later.

kjvbrt:
Nice :)

PR #79, opened 2024-04-24
Add cluster shape variables

dasphy:
Hi all,
Following PR #72 , more shape variables added for clusters.
These variables showed good separations for photon/pi0 identification.

@BrieucF , @kjvbrt 
Tagging @giovannimarchiori , @AlexisMlz , @gartrog 

Cheers, Tong


RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,22 +87,55 @@ StatusCode AugmentClustersFCCee::initialize()
...
     }
   }
   m_showerShapeHandle.put(showerShapeDecorations);
 
   return StatusCode::SUCCESS;
 }
 
+std::pair<std::vector<int>, std::vector<double>> MergeSumAndSort(std::vector<int>& A, std::vector<double>& B) {
+  std::unordered_map<int, double> elementSum;
+  // traverse vec A, merge the same elements (theta ID) in vec A and sum up the corresponding elements (E_cell) in vec B
```
giovannimarchiori, line 104: I would move the description of the method above its definition, and also add to the description that the method returns a pair of vectors with the merged cells (unique elements of vector A and sum of values of B for cells with same ID), sorted in ascending order of identifier (explain also why the sorting is needed)

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,22 +87,55 @@ StatusCode AugmentClustersFCCee::initialize()
       showerShapeDecorations.push_back(Form("energy_fraction_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("theta_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("phi_%s_layer_%d", detector, layer));
+      showerShapeDecorations.push_back(Form("width_theta_%s_layer_%d", detector, layer));
```
giovannimarchiori, line 90: the idea was that these decorations would be added conditionally on a boolean flag. can you add it?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,22 +87,55 @@ StatusCode AugmentClustersFCCee::initialize()
...
+      showerShapeDecorations.push_back(Form("Delta_E_2ndmax_min_%s_layer_%d", detector, layer));
+      showerShapeDecorations.push_back(Form("width_theta_3Bin_%s_layer_%d", detector, layer));
     }
   }
   m_showerShapeHandle.put(showerShapeDecorations);
 
   return StatusCode::SUCCESS;
 }
 
+std::pair<std::vector<int>, std::vector<double>> MergeSumAndSort(std::vector<int>& A, std::vector<double>& B) {
```
giovannimarchiori, line 102: A and B are not modified, right? So why not using a const modifier?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -120,38 +163,48 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
     sumEnLayer.assign(numLayersTotal, 0.);
     double phiMin = 9999.;
     double phiMax = -9999.;
+    int module_id_Min = 1536;
```
giovannimarchiori, line 166: this should not be hardcoded. It's a property of the detector geometry, I think you should be able to retrieve it from the geometry service - or even the readout

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -120,38 +163,48 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+    int module_id_Max = -1;
 
     // loop over each system/readout
     int startPositionToFill = 0;
     for (size_t k = 0; k < m_readoutNames.size(); k++)
     {
       if (k > 0)
       {
-        startPositionToFill += m_numLayers[k - 1];
+        startPositionToFill += m_numLayers[k - 1];  // only 2 sub detectors ?
```
giovannimarchiori, line 175: these shower shape variables for photon/pi0 discrimination are only useful for the ECAL.. does it make sense to compute and save them also for the HCAL?

giovannimarchiori: I would probably at least add a boolean flag to calculate or not these pi0/gamma shower shapes so that the user can turn them on only for ECAL-only clusters

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -161,6 +214,11 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
     debug() << "phiMin, phiMax : " << phiMin << " " << phiMax << endmsg;
     debug() << "Cluster is near phi=pi : " << isClusterPhiNearPi << endmsg;
 
+    bool isResetModuleID = false;
+    if (module_id_Max - module_id_Min > 1500)    isResetModuleID = true;
```
giovannimarchiori, line 218: again, not good to use hardcoded numbers...

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -182,33 +259,132 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
         double phi = v.Phi();
+
         // for clusters that are around the -pi<->pi transition, we want to avoid averaging
         // over phi values that might differ by 2pi. in that case, for cells with negative
         // phi we add two pi, so that we average phi values all close to pi
-        if (isClusterPhiNearPi && phi < 0.)
+        if (isClusterPhiNearPi && phi < 0.) {
           phi += TMath::TwoPi();
+        }
+        if (isResetModuleID && module_id > 1536/2) {
```
giovannimarchiori, line 290: hardcoded numbers!!!

dasphy: thanks, revising these issues

dasphy:
Thanks @giovannimarchiori . I pushed a new commit to fix these issues.
Cheers, Tong

dasphy:
Related PR in LAr_scripts:
https://github.com/BrieucF/LAr_scripts/pull/26

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +283,151 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+        if (vec_E_cell_layer[layer+startPositionToFill].empty()) {
+          vec_E_cell_layer[layer+startPositionToFill].push_back(0);
+          vec_theta_cell_layer[layer+startPositionToFill].push_back(0);
+          vec_module_cell_layer[layer+startPositionToFill].push_back(0);
+        }
+        auto result = MergeSumAndSort(vec_theta_cell_layer[layer+startPositionToFill], vec_E_cell_layer[layer+startPositionToFill]);
+        theta_E_pair.push_back(result);
+
+        // loop over theta IDs
+        for (size_t i = 0; i < theta_E_pair[layer+startPositionToFill].second.size(); i ++) {
```
giovannimarchiori, line 371: I think the part that finds the local maxima and the minimum between them is not behaving properly when there are some theta bins with 0 energy, since the corresponding cells are not attached to the cluster. One should take that into account

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +283,151 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
         double phi = v.Phi();
+
         // for clusters that are around the -pi<->pi transition, we want to avoid averaging
         // over phi values that might differ by 2pi. in that case, for cells with negative
         // phi we add two pi, so that we average phi values all close to pi
-        if (isClusterPhiNearPi && phi < 0.)
+        if (isClusterPhiNearPi && phi < 0.) {
           phi += TMath::TwoPi();
+        }
+        if (systemID == 4 && isResetModuleID && module_id > nModules_EMB/2) {
```
BrieucF, line 317: `mydetector.constantAsDouble("DetID_ECAL_Barrel")`

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +283,151 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
         if (m_thetaRecalcLayerWeights[k][layer]<0)
           sumThetaLayer[layer+startPositionToFill] += (eCell * theta);
         else
           sumThetaLayer[layer+startPositionToFill] += (weightLog * theta);
         sumWeightLayer[layer+startPositionToFill] += weightLog;
         sumPhiLayer[layer+startPositionToFill] += (eCell * phi);
-      }
+
+        // do pi0/photon shape var only for EMB
+        if (m_do_pi0_photon_shapeVar && systemID == 4) {
```
BrieucF, line 329: `mydetector.constantAsDouble("DetID_ECAL_Barrel");`

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +283,151 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+    std::vector<double> E_cell_Max(numLayersTotal, 0.);
+    std::vector<double> E_cell_secMax(numLayersTotal, 0.);
+    std::vector<double> E_cell_Max_theta(numLayersTotal, 0.);
+    std::vector<double> E_cell_secMax_theta(numLayersTotal, 0.);
+    std::vector<double> E_cell_Min(numLayersTotal, std::numeric_limits<double>::max());
+
+    startPositionToFill = 0;
+    for (size_t k = 0; k < m_readoutNames.size(); k++) {
+      if (!m_do_pi0_photon_shapeVar)    break;
+      if (m_systemIDs[k] != 4)    continue;    // do pi0/photon shape var only for EMB
```
BrieucF, line 357: `mydetector.constantAsDouble("DetID_ECAL_Barrel");`  Actually since it appears many times, you could just define a `systemID_EMB = mydetector.constantAsDouble("DetID_ECAL_Barrel");` once.

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -245,11 +454,64 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+
+        newCluster.addToShapeParameters(sumEnLayer[layer+startPositionToFill] / E);  // E fraction of layer
         newCluster.addToShapeParameters(sumThetaLayer[layer+startPositionToFill]);
         newCluster.addToShapeParameters(sumPhiLayer[layer+startPositionToFill]);
-      }
-    }
-  }
+
+        // do pi0/photon shape var only for EMB
+        if (m_do_pi0_photon_shapeVar && systemID == 4) {
```
BrieucF, line 463: See recurrent comment

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
```diff
@@ -10,17 +10,22 @@ class IGeoSvc;
 #include "GaudiKernel/Algorithm.h"
 #include "GaudiKernel/ToolHandle.h"
 
+// DD4HEP
+#include "DDSegmentation/Segmentation.h"
```
BrieucF, line 14: Is this needed here?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +283,151 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+        }
+        auto result = MergeSumAndSort(vec_theta_cell_layer[layer+startPositionToFill], vec_E_cell_layer[layer+startPositionToFill]);
+        theta_E_pair.push_back(result);
+
+        // loop over theta IDs
+        for (size_t i = 0; i < theta_E_pair[layer+startPositionToFill].second.size(); i ++) {
+          // find the local E maxima
+          // discard the first and last cell in theta
+          // local max: E_cell is greater than the 2 neighbors
+          if (i != 0 && i != (theta_E_pair[layer+startPositionToFill].second.size()-1) &&
```
giovannimarchiori, line 375: I would not discard the cells at the edges of the cluster - if the cell with theta = theta_min has energy greater than that of theta_min+1, then that's the one to keep. So I would replace with 

`          if (
            (i == 0 && theta_E_pair[layer+startPositionToFill].second[i] > theta_E_pair[layer+startPositionToFill].second[i+1]) ||
            (i == theta_E_pair[layer+startPositionToFill].second.size()-1 && theta_E_pair[layer+startPositionToFill].second[i] > theta_E_pair[layer+startPositionToFill].second[i-1]) ||
            (i != 0 && i != (theta_E_pair[layer+startPositionToFill].second.size()-1) && 
              theta_E_pair[layer+startPositionToFill].second[i] > theta_E_pair[layer+startPositionToFill].second[i-1] && 
              theta_E_pair[layer+startPositionToFill].second[i] > theta_E_pair[layer+startPositionToFill].second[i+1])
           ) {`

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -9,6 +9,10 @@
 // DD4hep
 #include "DD4hep/Detector.h"
 
+// k4geo
+#include "detectorSegmentations/FCCSWGridPhiTheta_k4geo.h"
```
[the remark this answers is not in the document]
dasphy: this was not used, can be removed

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,28 +91,89 @@ StatusCode AugmentClustersFCCee::initialize()
...
   const edm4hep::ClusterCollection *inClusters = m_inClusters.get();
-
   // create the new output collection
   edm4hep::ClusterCollection *outClusters = m_outClusters.createAndPut();
 
+  // get segmentation and nModules of EMB
+  int nModules_EMB = -1;
+  for (size_t k = 0; k < m_readoutNames.size(); k++) {
+    std::string readoutName = m_readoutNames[k];
+    if (readoutName != "ECalBarrelModuleThetaMerged")    continue;
```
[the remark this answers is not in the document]
dasphy: OK.

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
dasphy: OK, so will do the loop over cells instead

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
dasphy: it can be removed

BrieucF:
Hi @dasphy , any news on this?

dasphy:
Hi @BrieucF , thanks. More shape variables added, will push the updated code later today

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,28 +91,89 @@ StatusCode AugmentClustersFCCee::initialize()
...
 {
-
-
   // get the input collection with clusters
   const edm4hep::ClusterCollection *inClusters = m_inClusters.get();
-
   // create the new output collection
   edm4hep::ClusterCollection *outClusters = m_outClusters.createAndPut();
 
+  // get segmentation and nModules of EMB
```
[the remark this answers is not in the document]
dasphy: OK, moved to initialize()

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -77,28 +91,89 @@ StatusCode AugmentClustersFCCee::initialize()
...
+  int nModules_EMB = -1;
+  for (size_t k = 0; k < m_readoutNames.size(); k++) {
```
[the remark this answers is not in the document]
dasphy: checked that at the beginning of initialize()

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
dasphy: Filled zero energy cells in 1D theta-E and module-E profiles

dasphy:
Hi @BrieucF, @giovannimarchiori 
I improved some calculations, now the shape variable distributions look good to me

giovannimarchiori:
@dasphy can you put the plots in some slides and upload them to cernbox please?

giovannimarchiori:
@dasphy can you check the calculation of Ratio_E_max_2ndmax_vs_phi_EMB_layer_XX ? I see there are several NaN in the root files you produced, typically for innermost or outermost layers (0,1, 8, 9, 10...). This makes me think that when there is not enough energy in the layer the calculation fails and a NaN is returned (probably a 0 in the denominator of some ratio).

dasphy:
> @dasphy can you check the calculation of Ratio_E_max_2ndmax_vs_phi_EMB_layer_XX ? I see there are several NaN in the root files you produced, typically for innermost or outermost layers (0,1, 8, 9, 10...). This makes me think that when there is not enough energy in the layer the calculation fails and a NaN is returned (probably a 0 in the denominator of some ratio).

OK fixed. Used a wrong var name for nan check.
          if (E_cell_vs_phi_Max[layer+startPositionToFill] + E_cell_vs_phi_secMax[layer+startPositionToFill] == 0.)    Ratio_E = 1.;
=>
          if (E_cell_vs_phi_Max[layer+startPositionToFill] + E_cell_vs_phi_secMax[layer+startPositionToFill] == 0.)    Ratio_E_vs_phi = 1.;

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
```diff
@@ -80,6 +88,18 @@ class AugmentClustersFCCee : public Gaudi::Algorithm
   /// Name of the layer/cell field
   Gaudi::Property<std::vector<std::string>> m_layerFieldNames{
       this, "layerFieldNames", {"layer"}, "Identifiers of layers, corresponding to systemIDs"};
+  Gaudi::Property<std::vector<std::string>> m_thetaFieldNames{
+      this, "thetaFieldNames", {"theta"}, "Identifiers of theta, corresponding to systemIDs"};
+  Gaudi::Property<std::vector<std::string>> m_moduleFieldNames{
+      this, "moduleFieldNames", {"module"}, "Identifiers of module, corresponding to systemIDs"};
+  Gaudi::Property<bool> m_do_pi0_photon_shapeVar{
+      this, "do_pi0_photon_shapeVar", false, "Calculate shape variables for pi0/photon separation: E_ratio, Delta_E etc."};
```
giovannimarchiori, line 96: Hi @dasphy ,

does the code work for you if this property is set to false? In my local tests it crashes, can you please check?

giovannimarchiori:
Hi @dasphy , 
the E_fr_side_pm2_EMB_layer_N,  E_fr_side_pm3_EMB_layer_N, width_theta_3Bin_EMB_layer_N, width_theta_5Bin_EMB_layer_N, width_theta_7Bin_EMB_layer_N are 0 for all layers N that do not have strips.  I have verified that changing strip layer changes the variables that are 0. I think in the width calculation you're probably not accounting from the fact that in the non-strip layers the cells are bigger (thetaID increases by 4 among neighbours). Can you fix this?
Also, the list of variables being added is quite long, probably OK at the moment for R&D studies, then we should remember later to reduce it

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
dasphy: Hi @giovannimarchiori that works for me. In the RECO script I added "do_pi0_photon_shapeVar=True", to save the shape parameters:

   augmentCaloTopoClusters = AugmentClustersFCCee("augmentCaloTopoClusters",
                                                       inClusters=createTopoClusters.clusters.Path,
                                                       outClusters="Augmented" + createTopoClusters.clusters.Path,
                                                       systemIDs=[4],
                                                       systemNames=["EMB"],
                                                       numLayers=[ecalBarrelLayers],
                                                       readoutNames=[ecalBarrelReadoutName],
                                                       layerFieldNames=["layer"],
                                                       thetaRecalcWeights=[ecalBarrelThetaWeights],
                                                  do_pi0_photon_shapeVar=True,
                                                       OutputLevel=INFO)

dasphy:
Yes you are right @giovannimarchiori . In non-strip layers the theta_ID should be increased by 4, not 1. Will fix that.

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
giovannimarchiori: Hi @dasphy ,
yes, with do_pi0_photon_shapeVar=True the code runs, but as I wrote before, with do_pi0_photon_shapeVar=False, at least for me, there is a segmentation violation. Are you sure that for you with do_pi0_photon_shapeVar=False the code runs successfully?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -177,48 +360,261 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+    }
+
+    // local maxima of E vs phi (could be more than one) and the corresponding module
+    // note that code should be modified to handle periodicity of module/phi (and tested on particles in the direction of the 1536..0 module transition)
+    std::vector<std::pair<std::vector<int>, std::vector<double>>> module_E_pair;
+    std::vector<std::vector<double>> local_E_Max_vs_phi(numLayersTotal, std::vector<double>());
+    std::vector<std::vector<int>> local_E_Max_vs_phi_module(numLayersTotal, std::vector<int>());
+
+    startPositionToFill = 0;
+    for (size_t k = 0; k < m_readoutNames.size(); k++) {
```
giovannimarchiori, line 520: this for instance looks like a block of code that should be only executed if do_pi0_photon_shapeVar is true, otherwise the input vectors will be empty. This might explain the crashes (ReferenceError: attempt to access a null-pointer) observed when the algorithm is configured with do_pi0_photon_shapeVar = false

giovannimarchiori:
> Yes you are right @giovannimarchiori . In non-strip layers the theta_ID should be increased by 4, not 1. Will fix that.

What about the variables vs module? There we are merging every two, are you increasing module ID by 1 or by 2? And, needless to say, do not use hardcoded numbers but the variables encoding the number of merged strips or merged modules per layer

dasphy:
> > Yes you are right @giovannimarchiori . In non-strip layers the theta_ID should be increased by 4, not 1. Will fix that.
> 
> What about the variables vs module? There we are merging every two, are you increasing module ID by 1 or by 2? And, needless to say, do not use hardcoded numbers but the variables encoding the number of merged strips or merged modules per layer

what would this affect would be width_theta/phi and fr_side calculation. I didn't compute width_phi_nBins and fr_side_phi_nBins so for module variables should be OK. Also for width_module there should be no impact but anyway I will fix that and redo the production

kjvbrt:
Hi @dasphy, can you update this branch to the latest upstream, so the CI can test whether the PR passes through the nightlies?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.h
kjvbrt, line 1: Would be nice to use `edm4hep::labels::ShapeParameterNames` instead of `edm4hep::shapeParameterNames` :)

dasphy:
Thank you @kjvbrt , I updated the branch and switched to ```edm4hep::labels::ShapeParameterNames```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
         newCluster.addToShapeParameters(sumThetaLayer[layer+startPositionToFill]);
         newCluster.addToShapeParameters(sumPhiLayer[layer+startPositionToFill]);
-      }
-    }
-  }
+
+        // do pi0/photon shape var only for EMB
+        if (m_do_pi0_photon_shapeVar && systemID == systemID_EMB) {
+          double w_theta = sqrt(fabs(theta2_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill], 2)));
+          if (std::isnan(w_theta))    w_theta = 0.;
```
giovannimarchiori, line 632: w_theta should be nan only if sumEnLayer of this layer is zero, shouldn't it? It would be better to replace this with something that handles this more cleanly and doesn't hide other NaN that might results from bugs in the code
```suggestion
          double w_theta = (sumEnLayer[layer+startPositionToFill] == 0.) ? 0. : sqrt(fabs(theta2_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill], 2)));
```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+          if (w_theta > 40)    w_theta = 40;
```
giovannimarchiori, line 633: Why capping the variable? I would remove this line

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+          width_theta[layer+startPositionToFill] = w_theta;
+
+          double w_module = sqrt(fabs(module2_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill] - std::pow(module_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill], 2)));
+          if (std::isnan(w_module))    w_module = 0.;
+          if (w_module > 40)    w_module = 40;
```
giovannimarchiori, line 638: Why capping the variable? I would remove this line

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+
+        // do pi0/photon shape var only for EMB
+        if (m_do_pi0_photon_shapeVar && systemID == systemID_EMB) {
+          double w_theta = sqrt(fabs(theta2_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill], 2)));
+          if (std::isnan(w_theta))    w_theta = 0.;
+          if (w_theta > 40)    w_theta = 40;
+          width_theta[layer+startPositionToFill] = w_theta;
+
+          double w_module = sqrt(fabs(module2_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill] - std::pow(module_E_layer[layer+startPositionToFill] / sumEnLayer[layer+startPositionToFill], 2)));
+          if (std::isnan(w_module))    w_module = 0.;
```
giovannimarchiori, line 637: same comment as for w_theta

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -175,48 +359,242 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+
+        // do pi0/photon shape var only for EMB
+        if (m_do_pi0_photon_shapeVar && systemID == systemID_EMB) {
+          // E, theta_id, and module_id of cells in layer
+          vec_E_cell_layer[layer+startPositionToFill].push_back(eCell);
+          vec_theta_cell_layer[layer+startPositionToFill].push_back(theta_id);
+          vec_module_cell_layer[layer+startPositionToFill].push_back(module_id);
+          // sum them for width in theta/module calculation
+          theta2_E_layer[layer+startPositionToFill] += theta_id * theta_id * eCell;
+          theta_E_layer[layer+startPositionToFill] += theta_id * eCell;
```
giovannimarchiori, line 412: Since the theta response is better with logE weights, have you tried to calculate the w_theta from the mean theta and theta2 using the logE weights instead of the Ecell weights?

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -175,48 +359,242 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+        } else if (local_E_Max[layer+startPositionToFill].size() < 2) {
+          E_cell_Max[layer+startPositionToFill] = local_E_Max[layer+startPositionToFill][0];
+          E_cell_secMax[layer+startPositionToFill] = 0.;
+          E_cell_Max_theta[layer+startPositionToFill] = local_E_Max_theta[layer+startPositionToFill][0];
+          E_cell_secMax_theta[layer+startPositionToFill] = local_E_Max_theta[layer+startPositionToFill][0];
+          E_cell_Min[layer+startPositionToFill] = 0.;
+        } else {
+          std::vector<double> sortedVec = local_E_Max[layer+startPositionToFill];
+          // move the top 2 max to the beginning
+          std::partial_sort(sortedVec.begin(), sortedVec.begin()+2, sortedVec.end(), std::greater<double>());
```
giovannimarchiori, line 480: if you want only the first two, rather than the first three, you could  actually do
```suggestion
          std::partial_sort(sortedVec.begin(), sortedVec.begin()+1, sortedVec.end(), std::greater<double>());
```

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+            double _w_theta_9Bin = sqrt(fabs(theta2_E_9Bin / sum_E_9Bin - std::pow(theta_E_9Bin / sum_E_9Bin, 2)));
+
+            if (std::isnan(_w_theta_3Bin))    _w_theta_3Bin = 0.;
+            width_theta_3Bin[layer+startPositionToFill] = _w_theta_3Bin;
+            if (std::isnan(_w_theta_5Bin))    _w_theta_5Bin = 0.;
+            width_theta_5Bin[layer+startPositionToFill] = _w_theta_5Bin;
+            if (std::isnan(_w_theta_7Bin))    _w_theta_7Bin = 0.;
+            width_theta_7Bin[layer+startPositionToFill] = _w_theta_7Bin;
+            if (std::isnan(_w_theta_9Bin))    _w_theta_9Bin = 0.;
+            width_theta_9Bin[layer+startPositionToFill] = _w_theta_9Bin;
```
giovannimarchiori, line 775: does it ever happen that the energy sums in 3-5-7-9 bins are zero? This should not happen as long as you have a maximum energy cell with E>0... I'd rather check this or that the energy sums are zero, rather than correcting the nan shower shapes later (they could be nan due to some other bugs and we would not be able to figure out)

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+            theta_E_5Bin = theta_E_3Bin + theta_m2 * E_m2 + theta_p2 * E_p2;
+            theta2_E_7Bin = theta2_E_5Bin + theta_m3 * theta_m3 * E_m3 + theta_p3 * theta_p3 * E_p3;
+            theta_E_7Bin = theta_E_5Bin + theta_m3 * E_m3 + theta_p3 * E_p3;
+            theta2_E_9Bin = theta2_E_7Bin + theta_m4 * theta_m4 * E_m4 + theta_p4 * theta_p4 * E_p4;
+            theta_E_9Bin = theta_E_7Bin + theta_m4 * E_m4 + theta_p4 * E_p4;
+
+            double _w_theta_3Bin = sqrt(fabs(theta2_E_3Bin / sum_E_3Bin - std::pow(theta_E_3Bin / sum_E_3Bin, 2)));
+            double _w_theta_5Bin = sqrt(fabs(theta2_E_5Bin / sum_E_5Bin - std::pow(theta_E_5Bin / sum_E_5Bin, 2)));
+            double _w_theta_7Bin = sqrt(fabs(theta2_E_7Bin / sum_E_7Bin - std::pow(theta_E_7Bin / sum_E_7Bin, 2)));
+            double _w_theta_9Bin = sqrt(fabs(theta2_E_9Bin / sum_E_9Bin - std::pow(theta_E_9Bin / sum_E_9Bin, 2)));
```
giovannimarchiori, line 766: why the fabs? if you do these calculations only when there is energy in the layer (or sum_E_XXBin is >0) the fabs should not be needed

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -243,11 +621,195 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+            width_theta_7Bin[layer+startPositionToFill] = _w_theta_7Bin;
+            if (std::isnan(_w_theta_9Bin))    _w_theta_9Bin = 0.;
+            width_theta_9Bin[layer+startPositionToFill] = _w_theta_9Bin;
+
+            if (std::isnan(sum_E_5Bin / sum_E_3Bin))    E_fr_side_pm2[layer+startPositionToFill] = 0.;
+            else    E_fr_side_pm2[layer+startPositionToFill] = ((sum_E_5Bin/sum_E_3Bin-1.) > 10 ? 10. : (sum_E_5Bin/sum_E_3Bin-1.));
+            if (std::isnan(sum_E_7Bin / sum_E_3Bin))    E_fr_side_pm3[layer+startPositionToFill] = 0.;
+            else    E_fr_side_pm3[layer+startPositionToFill] = ((sum_E_7Bin/sum_E_3Bin-1.) > 10 ? 10. : (sum_E_7Bin/sum_E_3Bin-1.));
+            if (std::isnan(sum_E_9Bin / sum_E_3Bin))    E_fr_side_pm4[layer+startPositionToFill] = 0.;
+            else    E_fr_side_pm4[layer+startPositionToFill] = ((sum_E_9Bin/sum_E_3Bin-1.) > 10 ? 10. : (sum_E_9Bin/sum_E_3Bin-1.));
```
giovannimarchiori, line 782: again, all  these checks for nan should not be needed, and might hide potential issues in other parts of the code

giovannimarchiori:
Hi @dasphy , I finally managed to went through the full code once more after the latest changes. 
Overall I didn't spot any obvious bug. I think you should remove capping the values of some variables, and replace the checks for none as suggested. Also, it should be interesting to compare the discriminatning power if the theta width variables are calculating using linear or log energy weights (maybe you could add a boolean flag for that and implement is as an option). That's all from my side

dasphy:
> Hi @dasphy , I finally managed to went through the full code once more after the latest changes. Overall I didn't spot any obvious bug. I think you should remove capping the values of some variables, and replace the checks for none as suggested. Also, it should be interesting to compare the discriminatning power if the theta width variables are calculating using linear or log energy weights (maybe you could add a boolean flag for that and implement is as an option). That's all from my side

Thanks @giovannimarchiori for reviewing the code !
I applied some fixes accordingly. Will implement the logE weights in width calculation.

kjvbrt:
Hi @dasphy, do you plan to implement the logE weights in this PR?

dasphy:
Hi @kjvbrt @giovannimarchiori 
I implemented the logE weights in width theta calculation

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -664,6 +664,7 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
             double theta2_E_3Bin = 0.;
             double theta_E_3Bin = 0.;
             double sum_E_3Bin = 0.;
+            double sum_weightLog_3Bin = 0.;
```
giovannimarchiori, line 667: why declaring/calculating the wightlog variables here and not later inside the if (m_do_widthTheta_logE_weights) {..} block? Maybe the compiler is smart enough to optimise this code, but it makes anyway the code logic a bit harder to follow.

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -672,6 +673,7 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
             double theta2_E_5Bin = 0.;
             double theta_E_5Bin = 0.;
             double sum_E_5Bin = 0.;
+            double sum_weightLog_5Bin = 0.;
```
giovannimarchiori, line 676: Same comment as for L667

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -680,6 +682,7 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
             double theta2_E_7Bin = 0.;
             double theta_E_7Bin = 0.;
             double sum_E_7Bin = 0.;
+            double sum_weightLog_7Bin = 0.;
```
giovannimarchiori, line 685: Same comment as for L667

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -752,34 +756,71 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
+            double weightLog_m1 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_m1 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_m2 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_m2 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_m3 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_m3 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_m4 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_m4 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_p1 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_p1 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_p2 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_p2 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_p3 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_p3 / sumEnLayer[layer+startPositionToFill]));
+            double weightLog_p4 = std::max(0., m_thetaRecalcLayerWeights[k][layer] + log(E_p4 / sumEnLayer[layer+startPositionToFill]));
+
```
giovannimarchiori, line 768: Same comment as for L667

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -752,34 +756,71 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
             sum_E_3Bin = E_m1 + E_cell_Max[layer+startPositionToFill] + E_p1;
             sum_E_5Bin = sum_E_3Bin + E_m2 + E_p2;
             sum_E_7Bin = sum_E_5Bin + E_m3 + E_p3;
             sum_E_9Bin = sum_E_7Bin + E_m4 + E_p4;
+            sum_weightLog_3Bin = weightLog_E_max + weightLog_m1 + weightLog_p1;
+            sum_weightLog_5Bin = sum_weightLog_3Bin + weightLog_m2 + weightLog_p2;
+            sum_weightLog_7Bin = sum_weightLog_5Bin + weightLog_m3 + weightLog_p3;
+            sum_weightLog_9Bin = sum_weightLog_7Bin + weightLog_m4 + weightLog_p4;
```
giovannimarchiori, line 776: Same comment as for L667.
And the previous 4 lines (sum_E_3bin = ...) could go inside the else { .. }. block of L787

kjvbrt:
Nice, merging...

PR #80, opened 2024-05-08
Adding tools for jet clustering for cluster and truth particles

jroloff:
This PR adds two algorithms for jet clustering: one for calorimeter clusters, and another for truth jets. Several options are included for the jet radius, jet input name, and more.

### DIFF
--- a/RecCalorimeter/CMakeLists.txt
+++ b/RecCalorimeter/CMakeLists.txt
@@ -17,6 +17,7 @@ gaudi_add_module(k4RecCalorimeterPlugins
                       ROOT::Hist
                       k4geo::detectorSegmentations
                       k4geo::detectorCommon
+                      ${FASTJET_LIBRARIES}
                       )
 
 install(TARGETS k4RecCalorimeterPlugins
@@ -25,6 +26,8 @@ install(TARGETS k4RecCalorimeterPlugins
   LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
   COMPONENT dev)
 
+include_directories(${FASTJET_INCLUDE_DIRS})
+link_directories(${FASTJET_LIBRARIES})
 
 include(CTest)
 
--- a/RecCalorimeter/CMakeLists.txt
+++ b/RecCalorimeter/CMakeLists.txt
@@ -38,12 +38,13 @@ add_test(NAME FCC_createJets
   COMMAND k4run RecCalorimeter/tests/options/runJetReco.py
          WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 )
+set_test_env(FCC_createJets)
 
 add_test(NAME FCC_createTruthJets
   COMMAND k4run RecCalorimeter/tests/options/runTruthJetReco.py
          WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 )
-
+set_test_env(FCC_createTruthJets)
 
 #install(DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/tests/options DESTINATION ${CMAKE_INSTALL_DATADIR}/${CMAKE_PROJECT_NAME}/Reconstruction/RecCalorimeter)
 #
--- a/RecCalorimeter/CMakeLists.txt
+++ b/RecCalorimeter/CMakeLists.txt
@@ -3,9 +3,10 @@
 ################################################################################
 
 
-file(GLOB _module_sources src/components/*.cpp)
+file(GLOB _module_sources src/components/*.cpp )
+
 gaudi_add_module(k4RecCalorimeterPlugins
-                 SOURCES ${_module_sources}
+                 SOURCES ${_module_sources} 
                  LINK k4FWCore::k4FWCore
                       k4FWCore::k4Interface
                       Gaudi::GaudiAlgLib
@@ -26,7 +27,9 @@ install(TARGETS k4RecCalorimeterPlugins
   LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
   COMPONENT dev)
 
-include_directories(${FASTJET_INCLUDE_DIRS})
+
+
+include_directories(${FASTJET_INCLUDE_DIRS} include/RecCalorimeter)
 link_directories(${FASTJET_LIBRARIES})
 
 include(CTest)
--- a/RecCalorimeter/CMakeLists.txt
+++ b/RecCalorimeter/CMakeLists.txt
@@ -31,6 +31,17 @@ link_directories(${FASTJET_LIBRARIES})
 
 include(CTest)
 
+add_test(NAME FCC_createJets
+  COMMAND k4run RecCalorimeter/tests/options/runJetReco.py
+         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+)
+
+add_test(NAME FCC_createTruthJets
+  COMMAND k4run RecCalorimeter/tests/options/runTruthJetReco.py
+         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+)
+
+
 #install(DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/tests/options DESTINATION ${CMAKE_INSTALL_DATADIR}/${CMAKE_PROJECT_NAME}/Reconstruction/RecCalorimeter)
 #
 #gaudi_add_test(genJetClustering
--- a/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
+++ b/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
@@ -1,5 +1,7 @@
-#ifndef RECCALORIMETER_CLUSTERJET_H
-#define RECCALORIMETER_CLUSTERJET_H
+//#ifndef RECCALORIMETER_CLUSTERJET_H
+//#define RECCALORIMETER_CLUSTERJET_H
+#ifndef k4RecCalorimeterPlugins_CLUSTERJET_H
+#define k4RecCalorimeterPlugins_CLUSTERJET_H
 
 // std
 #include <map>
@@ -30,7 +32,7 @@
 #include "edm4hep/ClusterCollection.h"
 
 
-
+namespace k4::recCalo{
 
 class ClusterInfo : public fastjet::PseudoJet::UserInfoBase{
  public:
@@ -76,4 +78,5 @@ private:
 
 };
 
+}
 #endif /* RECCALORIMETER_CLUSTERJET_H */
--- a/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
+++ b/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
@@ -24,15 +24,27 @@ class ClusterInfo : public fastjet::PseudoJet::UserInfoBase{
 };
 
 
-
 /** @class ClusterJet k4RecCalorimeter/RecCalorimeter/src/components/ClusterJet.h
  *
+ *  Helper class for reconstructing jets, used to avoid code duplication in the classes that require specific input types.
+ *  It takes a vector of fastjet::PseudoJets as inputs, and returns the a vector of pseudojets after clustering.
+ *  
+ *  JetAlg: A string corresponding to the jet algorithm for clustering
+ *  JetRadius: The radius of the jets being clustered
+ *  MinPt: The pT threshold below which jets are ignored
+ *  isExclusiveClustering: 1 if jets should use an exclusive clustering, 0 otherwise
+ *  clusterArgs: Any other clustering arguments, such as the number of jets for exclusive clustering. 
+ *               This is not used everywhere, so check the code to make sure it is implemented for your use-case.
+ *
  *  @author Jennifer Roloff
+ *  @date 2024-7
  */
+
+
 
 class ClusterJet {
 public:
-  ClusterJet(std::string jetAlg, double jetRadius, int clustering = 0, double minPt = 0, int clusterArgs = 0);
+  ClusterJet(std::string jetAlg, double jetRadius, int isExclusiveClustering = 0, double minPt = 0, int clusterArgs = 0);
   StatusCode initialize();
   ~ClusterJet(){if(m_clustSeq) delete m_clustSeq;}
 
@@ -53,7 +65,7 @@ private:
 
   std::string m_jetAlg = "antikt";
   double m_jetRadius = 0.4;
-  int m_clustering = 0; // Inclusive clustering by default
+  int m_isExclusiveClustering = 0; // Inclusive clustering by default
   double m_minPt = 10; // Only relevant for inclusive clustering
   int m_njets = 0; // Only relevant for exclusive clustering
 
--- a/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
+++ b/RecCalorimeter/include/RecCalorimeter/ClusterJet.h
@@ -1,32 +1,15 @@
-//#ifndef RECCALORIMETER_CLUSTERJET_H
-//#define RECCALORIMETER_CLUSTERJET_H
-#ifndef k4RecCalorimeterPlugins_CLUSTERJET_H
-#define k4RecCalorimeterPlugins_CLUSTERJET_H
+#ifndef RECCALORIMETER_CLUSTERJET_H
+#define RECCALORIMETER_CLUSTERJET_H
+//#ifndef k4RecCalorimeterPlugins_CLUSTERJET_H
+//#define k4RecCalorimeterPlugins_CLUSTERJET_H
 
 // std
 #include <map>
 
-/*
-// Gaudi
-#include "GaudiKernel/ToolHandle.h"
-#include "GaudiKernel/MsgStream.h"
-#include "GaudiKernel/StatusCode.h"
-#include "GaudiAlg/Transformer.h"
-#include "k4FWCore/BaseClass.h"
-
-// k4FWCore
-#include "k4FWCore/DataHandle.h"
-
-#include "edm4hep/ReconstructedParticleCollection.h"
-#include "edm4hep/MCParticleCollection.h"
-*/
-
-#include "fastjet/JetDefinition.hh"
-
-
 // Gaudi
 #include "GaudiAlg/GaudiAlgorithm.h"
 
+#include "fastjet/JetDefinition.hh"
 
 // EDM4hep
 #include "edm4hep/ClusterCollection.h"
@@ -53,8 +36,9 @@ class ClusterJet {
 public:
   ClusterJet(std::string jetAlg, double jetRadius, int clustering = 0, double minPt = 0, int clusterArgs = 0);
   StatusCode initialize();
+  ~ClusterJet(){if(m_clustSeq) delete m_clustSeq;}
 
-  std::vector<fastjet::PseudoJet> cluster(const   std::vector<fastjet::PseudoJet> clustersPJ) const;
+  std::vector<fastjet::PseudoJet> cluster(const   std::vector<fastjet::PseudoJet> clustersPJ);
 
 private:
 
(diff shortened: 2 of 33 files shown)


RecFCCeeCalorimeter/src/components/JetFCCee.cpp
```diff
@@ -0,0 +1,89 @@
...
+JetFCCee::JetFCCee(const std::string& name, ISvcLocator* svcLoc) : GaudiAlgorithm(name, svcLoc) {
+  declareProperty("TopoClusterInput", m_inputClusters, "Handle for input cluster collection");
+  declareProperty("JetOutput", m_jetCollection, "Handle for output jet collection");
+  declareProperty("JetAlg", m_jetAlg, "Name of jet clustering algorithm");
+  declareProperty("JetRadius", m_jetRadius, "Jet clustering radius");
+  declareProperty("MinPt", m_minPt, "Minimum pT for saved jets");
+}
+
+
+StatusCode JetFCCee::initialize() {
```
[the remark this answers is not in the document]
jroloff: Yes, I've added this for both algorithms.

RecFCCeeCalorimeter/CMakeLists.txt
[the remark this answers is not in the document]
jroloff: Yes, I have added two separate ones for the Calo and Truth jets.

RecFCCeeCalorimeter/src/components/TruthJetFCCee.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  fastjet::JetDefinition* jetDef = new fastjet::JetDefinition( m_jetAlgMap[m_jetAlg], m_jetRadius);
+
+
+  std::vector<fastjet::PseudoJet> clustersPJ;
+  int i=0;
+  for(auto cluster: *inParticles){
+    if(cluster.isCreatedInSimulation()) continue;
+    if(cluster.getGeneratorStatus() != 1) continue;
+    // TODO: decide if muons should be included in the truth jet clustering
```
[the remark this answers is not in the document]
jroloff: Yes, I've added a new class for this.

jroloff:
Hi @BrieucF -- I think I've addressed the comments you made. I've started the functionality for exclusive jets, so it should be possible, but the default is to use inclusive jets for now. Let me know if you have any other comments, since this has changed quite a bit.

There are still some features I'd like to add, but if there aren't any problems with this version, I might prefer to make a git issue for any new features so we can merge this in the meantime.

jroloff:
Pinging @BrieucF to see if you have any more comments for this MR.

kjvbrt:
Adding @jroloff to the HEP-FCC GH organization in order for the CI to start automatically.

RecCalorimeter/src/components/ClusterJet.h
kjvbrt, line 1: Would it be possible to move this into separate subdirectory of `k4RecCalorimeter`, similarly to [SimG4Common]https://github.com/HEP-FCC/k4SimGeant4/tree/main/SimG4Common)?
This directory contains only Gaudi algorithms...

Also, enclose your classes in some namespace, for example `k4::recCalo` (I have no preference on the name :))

kjvbrt: The classes which are not Gaudi algorithms.

RecCalorimeter/src/components/CreateTruthJet.cpp
```diff
@@ -0,0 +1,68 @@
...
+    for(auto constit : constits){
+      edm4hep::MutableReconstructedParticle jetInput;
+      jetInput.setMomentum(edm4hep::Vector3f(constit.px(), constit.py(), constit.pz()));
+      jetInput.setEnergy(constit.e());
+      jetInput.setMass(constit.m());
+
+      int index = constit.user_info<ClusterInfo>().index();
+      jetInput.setPDG((input)[index].getPDG());
+
+      jet.addToParticles(jetInput);
```
BrieucF, line 58: Is'nt this `jetInput` lost in the output file? One should probably store also this new collection of ReconstructedParticle in the output rootfile. Here is an example of `MultiTransformer`: https://github.com/key4hep/k4FWCore/blob/main/test/k4FWCoreTest/src/components/ExampleFunctionalTransformerMultiple.cpp#L51.

I agree that this is not perfect, having a relation between "genJet" and MCParticle would feel more natural but for now there is no such thing AFAIK.

RecFCCeeCalorimeter/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 # ONNX DEPENDENCIES
 # includes
-include_directories("${ONNXRUNTIME_INCLUDE_DIRS}")
+include_directories(${ONNXRUNTIME_INCLUDE_DIRS})
```
BrieucF, line 7: Is this intended?

RecCalorimeter/src/components/ClusterJet.h
```diff
@@ -0,0 +1,79 @@
...
+#include "GaudiKernel/StatusCode.h"
+#include "GaudiAlg/Transformer.h"
+#include "k4FWCore/BaseClass.h"
+
+// k4FWCore
+#include "k4FWCore/DataHandle.h"
+
+#include "edm4hep/ReconstructedParticleCollection.h"
+#include "edm4hep/MCParticleCollection.h"
+*/
```
BrieucF, line 20: Should be removed if not needed

RecCalorimeter/src/components/ClusterJet.h
```diff
@@ -0,0 +1,79 @@
...
+ ClusterInfo(const int & index) : _index(index){}
+  int index() const{ return _index; }
+ protected:
+  int _index;
+};
+
+
+
+/** @class ClusterJet k4RecCalorimeter/RecCalorimeter/src/components/ClusterJet.h
+ *
```
BrieucF, line 46: Can you describe the tool here? Adding what an end-user should know to properly configure/use it without having to read the code?

BrieucF (commented):
Wonderful! Thanks Jennifer. I have a few comments but it looks good! 

`CreateTruthJet` and `FilterTruthParticles` should probably not live in k4RecCalorimeter (k4Gen would be a more natural place) but I would understand if you prefer to deal with this in another PR or to leave it on us for a future code re-organization.

RecCalorimeter/src/components/CreateTruthJet.cpp
BrieucF: Another suggestion (thanks @tmadlener) is to use Association instead of relations (the difference is that associations are separated object with pointers to pairs of objects, used e.g. to link RECO to GEN quantities that should therefore not be used at analysis level). This will 1) avoid loosing information when casting MCParticles to  ReconstructedParticles  2) avoid duplicating content on disk since the associations are pointers. 

One should use `MCRecoParticleAssociation`:
```
edm4hep::MCRecoParticleAssociation:
    Description: "Association between a ReconstructedParticle and the corresponding MCParticle"
    Author: "EDM4hep authors"
    Members:
      - float weight                        // weight of this association
    OneToOneRelations:
     - edm4hep::ReconstructedParticle  rec  // reference to the reconstructed particle
     - edm4hep::MCParticle sim              // reference to the Monte-Carlo particle
``` 

After defining the association object, you can set it with
```
        relation.setSim(MyMCParticle);
        relation.setRec(MyRecoParticle);
```

RecCalorimeter/src/components/FilterTruthParticles.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  declareProperty("SaveMuons", m_useMuons, "Bool for if muons are stored");
+  declareProperty("MinPt", m_minPt, "Minimum pT for filtered particles");
+}
+
+
+
+colltype_out  FilterTruthParticles::operator()(const colltype_in& input) const{
+
+  edm4hep::MCParticleCollection filteredParticles = edm4hep::MCParticleCollection();
```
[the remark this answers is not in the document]
jroloff: I expect that we will want to create files where we only have the filtered particles, so I think it would be best to leave it as-is.

RecFCCeeCalorimeter/CMakeLists.txt
jroloff: No, I'll revert this -- this was for some local changes that I was testing.

RecCalorimeter/src/components/CreateCaloJet.h
```diff
@@ -0,0 +1,70 @@
...
+
+private:
+
+
+  double m_minPt = 10;
+  double m_jetRadius = 0.4;
+  std::string m_jetAlg = "antikt";
+
+  // Map between jet algorithm names and the actual jet clustering algorithm
+  const std::map<const std::string, const fastjet::JetAlgorithm> m_jetAlgMap = {
```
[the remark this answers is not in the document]
jroloff: Done

RecCalorimeter/src/components/FilterTruthParticles.cpp
```diff
@@ -0,0 +1,56 @@
...
+    /*
+    edm4hep::MutableMCParticle filteredParticle = edm4hep::MutableMCParticle(particle);
+    filteredParticle.setPDG(particle.getPDG());
+    filteredParticle.setGeneratorStatus(particle.getGeneratorStatus());
+    filteredParticle.setSimulatorStatus(particle.getSimulatorStatus());
+    filteredParticle.setCharge(particle.getCharge());
+    filteredParticle.setTime(particle.getTime());
+    filteredParticle.setMass(particle.getMass());
+    filteredParticles.push_back(filteredParticle);
+    */
```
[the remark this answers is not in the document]
jroloff: Yes, done

RecCalorimeter/src/components/CreateCaloJet.h
```diff
@@ -0,0 +1,70 @@
...
+
+
+/** @class CreateCaloJet k4RecCalorimeter/RecFCCeeCalorimeter/src/components/CreateCaloJet.h
+ *
+ *  @author Jennifer Roloff
+ */
+
+
+using colltype_in  = edm4hep::ClusterCollection;
+using colltype_out = edm4hep::ReconstructedParticleCollection;
```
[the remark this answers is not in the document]
jroloff: Agreed that this makes more sense -- I had just been copying some other code, and hadn't thought too deeply about why it had been written that way. Changed now.

RecCalorimeter/src/components/FilterTruthParticles.h
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+// Datamodel
+namespace edm4hep {
+class MCParticleCollection;
+}
+
+
+/** @class FilterTruthParticles k4RecCalorimeter/RecCalorimeter/src/components/FilterTruthParticles.h
```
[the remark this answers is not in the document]
jroloff: Done

RecCalorimeter/src/components/ClusterJet.h
jroloff: Done

RecCalorimeter/src/components/CreateTruthJet.cpp
jroloff: I might be misunderstanding this, but can these associations be used for a one-to-many map, or just a one-to-one relation? It definitely seems better than what I currently have, but it's not obvious to me how this should work in my particular case.

BrieucF: Hi Jennifer, very legit question :-) This may change in the future but for now, each entry in the association collection has a single pair. The way to store a "one to many" is therefore to add multiple entries, all having the same `setRec(MyRecoParticle)` but different `setSim(MyMCParticle)`. Storage wise it is ok. CPU wise we loose a bit at analysis stage, where retrieving all MCParticles associated to a GenJet will be done with:

```
for (const auto& assoc : associationColl) {
  if (assoc.getRec() == MyRecoParticle) {
    theMCParticle = rel.getSim();
  }
}
```

RecCalorimeter/CMakeLists.txt
```diff
@@ -25,9 +26,22 @@ install(TARGETS k4RecCalorimeterPlugins
...
 
+include_directories(${FASTJET_INCLUDE_DIRS})
+link_directories(${FASTJET_LIBRARIES})
 
 include(CTest)
 
+add_test(NAME FCC_createJets
+  COMMAND k4run RecCalorimeter/tests/options/runJetReco.py
+         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+)
```
kjvbrt, line 40: You need to set environment for your test, like so:
```
set_test_env(FCC_createJets)
```

RecCalorimeter/CMakeLists.txt
```diff
@@ -25,9 +26,22 @@ install(TARGETS k4RecCalorimeterPlugins
...
+
+add_test(NAME FCC_createTruthJets
+  COMMAND k4run RecCalorimeter/tests/options/runTruthJetReco.py
+         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+)
```
kjvbrt, line 45: Here as well

RecCalorimeter/src/components/FilterTruthParticles.cpp
kjvbrt, line 1: There already is a particle filtering tool in `k4Gen`:
https://github.com/HEP-FCC/k4Gen/blob/main/k4Gen/src/components/GenParticleFilter.h

Can you expand that one with your pT cut?

RecCalorimeter/src/components/FilterTruthParticles.cpp
```diff
@@ -0,0 +1,46 @@
...
+  edm4hep::MCParticleCollection filteredParticles = edm4hep::MCParticleCollection();
+
+  for(auto particle: input){
+
+    // Only want to include final state particles (before GEANT simulation)
+    if(particle.isCreatedInSimulation()) continue;
+    if(particle.getGeneratorStatus() != 1) continue;
+
+    // Sometimes we want to ignore muons for jet reconstruction
+    if(m_useMuons && std::abs(particle.getPDG()) == 13) continue;
```
BrieucF, line 31: If you set `m_useMuons` to `true`, you actually skip muons in this logic, no?

RecCalorimeter/src/components/ClusterJet.h
BrieucF: @kjvbrt where you suggesting to also move ClusterJet.cpp?

RecCalorimeter/include/RecCalorimeter/ClusterJet.h
```diff
@@ -0,0 +1,65 @@
...
+ ClusterInfo(const int & index) : _index(index){}
+  int index() const{ return _index; }
+ protected:
+  int _index;
+};
+
+
+
+/** @class ClusterJet k4RecCalorimeter/RecCalorimeter/src/components/ClusterJet.h
+ *
```
BrieucF, line 29: Can you also add a description of the class here?

RecCalorimeter/include/RecCalorimeter/ClusterJet.h
```diff
@@ -0,0 +1,65 @@
...
+ *  @author Jennifer Roloff
+ */
+
+class ClusterJet {
+public:
+  ClusterJet(std::string jetAlg, double jetRadius, int clustering = 0, double minPt = 0, int clusterArgs = 0);
```
BrieucF, line 35: Either here or in the main description above, can you specify that `int clustering = 0` refers to exclusive/inclusive clustering? This is not easy to guess

BrieucF:
Thanks you very much for addressing the comments! After implementing the few remaining minor changes and fixing the tests, this is good to go!

kjvbrt:
Hi, the failing tests `FCCeeLAr_simRecoAllegroV3` and `FCCeeLAr_reproduceSegfault` are already fixed in the main.

RecCalorimeter/src/components/FilterTruthParticles.cpp
jroloff: Good catch, I will flip that logic.

RecCalorimeter/include/RecCalorimeter/ClusterJet.h
jroloff: I've changed the variable name to match the other tools, which also makes its definition clearer.

RecCalorimeter/src/components/CreateTruthJet.cpp
jroloff: Also, I guess this will only work for storing particles for truth jets, not for the calo jets? I'd prefer using the same solution for truth and calo jets, since I think this would make things very confusing otherwise.

BrieucF: The two solutions are already somewhat different (OneToManyRelations to Cluster vs to Particle) and storing a GEN level information into a RECO object can also be considered very confusing to some extent but I indeed see the 'pragmatic' advantage to be able to retrieve this information based on relation (and not association) in both cases. SO if this is the solution you prefer, let's go for it!

But can you confirm that the objects added with `jet.addToParticles(jetInput);` are accessible from the output rootfile?

jroloff: Yes, I ran a quick test in python where I printed out the energies of the clusters in a jet, and it ran without any problem. For reference, here is the code:

`from podio import root_io

podio_reader = root_io.Reader('output_fullCalo_SimAndDigi.root')

for event in podio_reader.get("events"):
    jets = event.get("Jets")
    for jet in jets:
      clusters = jet.getClusters()
      for cluster in clusters:
        print (cluster.getEnergy())`

So, I think this solution is sufficient for now, though it would be good to eventually have a solution that links the objects instead.

RecCalorimeter/src/components/FilterTruthParticles.cpp
jroloff: Yes, done

RecCalorimeter/src/components/CreateTruthJet.cpp
BrieucF: Ok for the clusters attached to `CaloJets` but my question was about the `truthJets`: can we access the ReconstructedParticles that created it? By reading the code, it looks like we are missing the saving of a collection containing the `jetInput` in the output rootfile.

jroloff: Sorry for the confusion, and you're right that it was broken for the truth jets. I've switched to using the associations for the truth jets, but not for the reco jets for now.

BrieucF: Great! I think the reco jets are fine with the oneToMany association to Clusters. 

But for the truthJets I think we should go for a mutlitransformer since we now have multiple output (not sure what will happen in multi-threaded environment in the current situation). The main change is that `TruthJetParticleAssociation` should appear as a `KeyValues`. Please find an example here: https://github.com/key4hep/k4FWCore/blob/main/test/k4FWCoreTest/src/components/ExampleFunctionalTransformerMultiple.cpp#L51


I tried to run `RecCalorimeter/tests/options/runTruthJetReco.py`, adding

```
################ Output
from Configurables import PodioOutput
out = PodioOutput("out")
out.outputCommands = ["keep *"]
out.filename = "output_truthJets.root"
ApplicationMgr().TopAlg += [out]
```

but the truthJet collection is empty (and the association as well). Could we adapt the test so that it actually produces some truthJets?

RecCalorimeter/src/components/ClusterJet.h
kjvbrt: Hi, `ClusterJet.cpp` should be also moved out of `src/components`

BrieucF:
Hi Jennifer, to avoid bothering you with more technicalities (you already did a great share), we will merge as is and refactor a few things in a later PR.

Thanks again!

RecCalorimeter/src/components/CreateTruthJet.cpp
jroloff: I've made both of these changes. I haven't added an output root file to the test, but I added it in temporarily to check that it produced truth jets, and that I was able to read the associations.

PR #85, opened 2024-06-24
Implement an algorithm for photon/pi0 discrimination based on input features and a trained model in ONNX format

giovannimarchiori:
The code in this PR provides a Gaudi algorithm that 
- reads the list of input features from a JSON file and the trained BDT model from an ONNX file
- for each cluster, reads the input features from the shapeParameters vector (if available), runs the inference, and saves the photon probability as an additional shape parameter of the new output cluster collection
It works with ONNX files created with either XGBoost and LightGBM (tested with both).
Could probably work also with different models based on features (DNN..) 

Uses as inputs the shower shapes calculated in https://github.com/HEP-FCC/k4RecCalorimeter/pull/79

### DIFF
--- a/RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
+++ b/RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
@@ -19,7 +19,7 @@ Ort::Value vec_to_tensor(std::vector<T> &data, const std::vector<std::int64_t> &
 }
 
 PhotonIDTool::PhotonIDTool(const std::string &name,
-			   ISvcLocator *svcLoc)
+                           ISvcLocator *svcLoc)
     : Gaudi::Algorithm(name, svcLoc)
 {
   declareProperty("inClusters", m_inClusters, "Input cluster collection");
@@ -62,32 +62,32 @@ StatusCode PhotonIDTool::initialize()
       // otherwise, we will use the energy of the cluster object
       auto it = std::find(shapeParameters.begin(), shapeParameters.end(), "rawE");
       if (it != shapeParameters.end())
-	{
-	  int position = std::distance(shapeParameters.begin(), it);
-	  m_inputPositionsInShapeParameters.push_back(position);
-	  info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
-	}
+      {
+        int position = std::distance(shapeParameters.begin(), it);
+        m_inputPositionsInShapeParameters.push_back(position);
+        info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
+      }
       else {
-	m_inputPositionsInShapeParameters.push_back(-1);
+        m_inputPositionsInShapeParameters.push_back(-1);
       }
     }
     else {
-      // for the other features, check if they are in the shape paramaters
+      // for the other features, check if they are in the shape parameters
       auto it = std::find(shapeParameters.begin(), shapeParameters.end(), feature);
       if (it != shapeParameters.end())
-	{
-	  int position = std::distance(shapeParameters.begin(), it);
-	  m_inputPositionsInShapeParameters.push_back(position);
-	  info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
-	}
+      {
+        int position = std::distance(shapeParameters.begin(), it);
+        m_inputPositionsInShapeParameters.push_back(position);
+        info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
+      }
       else
-	{
-	  // at least one of the inputs of the MVA was not found in the shapeParameters
-	  // so we can stop checking the others 
-	  m_inputPositionsInShapeParameters.clear();
-	  error() << "Feature " << feature << " not found, aborting..." << endmsg;
-	  return StatusCode::FAILURE;
-	}
+      {
+        // at least one of the inputs of the MVA was not found in the shapeParameters
+        // so we can stop checking the others
+        m_inputPositionsInShapeParameters.clear();
+        error() << "Feature " << feature << " not found, aborting..." << endmsg;
+        return StatusCode::FAILURE;
+      }
     }
   }
     
@@ -156,7 +156,7 @@ edm4hep::ClusterCollection *PhotonIDTool::initializeOutputClusters(
 }
 
 StatusCode PhotonIDTool::readMVAFiles(const std::string& mvaInputsFileName,
-				      const std::string& mvaModelFileName)
+                                      const std::string& mvaModelFileName)
 {
   // 1. read the file with the list of input features
   // TODO: figure how to save BDT to ONNX such that input is
@@ -289,7 +289,7 @@ StatusCode PhotonIDTool::readMVAFiles(const std::string& mvaInputsFileName,
 }
 
 StatusCode PhotonIDTool::applyMVAtoClusters(const edm4hep::ClusterCollection *inClusters,
-					    edm4hep::ClusterCollection *outClusters) const
+                                            edm4hep::ClusterCollection *outClusters) const
 {
   size_t numShapeVars = m_internal_input_names.size();
   std::vector<float> mvaInputs(numShapeVars);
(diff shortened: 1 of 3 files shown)


RecFCCeeCalorimeter/src/components/PhotonIDTool.h
```diff
@@ -0,0 +1,116 @@
...
+
+  /// Handle for output calorimeter clusters collection
+  mutable DataHandle<edm4hep::ClusterCollection> m_outClusters {
+    "outClusters", Gaudi::DataHandle::Writer, this
+  };
+
+  /// Handles for the cluster shower shape metadata to read and to write
+  MetaDataHandle<std::vector<std::string>> m_inShapeParameterHandle{
+    m_inClusters,
+    edm4hep::shapeParameterNames,
```
kjvbrt, line 88: Can you use `edm4hep::labels::ShapeParameterNames` instead?

giovannimarchiori: done

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -0,0 +1,352 @@
...
+  return outClusters;
+}
+
+StatusCode PhotonIDTool::readMVAFiles(const std::string& mvaInputsFileName,
+				      const std::string& mvaModelFileName)
+{
+  // 1. read the file with the list of input features
+  // TODO: figure how to save BDT to ONNX such that input is
+  // the list of input features with their names, rather
+  // than single tensor "X"
```
kjvbrt, line 164: Can you use `nlohmann_json` instead of parsing the JSON manually? It is a dependency anyway (make it explicit in the CMake file). To keep ability to have comments:
https://json.nlohmann.me/features/comments/

Also, why are you splitting part of the configuration outside of the steering file?
Do you have that many shape parameters?

If this is just temporary solution, ignore this comment.

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -0,0 +1,352 @@
...
+      // otherwise, we will use the energy of the cluster object
+      auto it = std::find(shapeParameters.begin(), shapeParameters.end(), "rawE");
+      if (it != shapeParameters.end())
+	{
+	  int position = std::distance(shapeParameters.begin(), it);
+	  m_inputPositionsInShapeParameters.push_back(position);
+	  info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
+	}
+      else {
+	m_inputPositionsInShapeParameters.push_back(-1);
```
kjvbrt, line 71: `m_inputPositionsInShapeParameters` is `unsigned short int`

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: Hi @kjvbrt, the number of shape parameters can exceed 100.
What are you referring to in particular when you write "splitting part of the configuration outside of the steering file"? Do you mean, putting the names of the shape variables in the json rather than in the .py that steers k4run? Or something else?
About nlohmann_json, I thought it was an overkill for just parsing a json that contains a list of strings (could have been a CSV). But OK, in the end I can probably use that, so that in the json I will also save the parameters used at training time which is always useful to record somewhere

kjvbrt: > What are you referring to in particular when you write "splitting part of the configuration outside of the steering file"? Do you mean, putting the names of the shape variables in the json rather than in the .py that steers k4run? Or something else?

Yes, this is what I mean, I would like to keep all configurations needed for running in one file, if possible. I agree that 100 parameters is a bit too much for the k4run steering file :)

> About nlohmann_json, I thought it was an overkill for just parsing a json that contains a list of strings (could have been a CSV). But OK, in the end I can probably use that, so that in the json I will also save the parameters used at training time which is always useful to record somewhere

I would not suggest using it if it was not already a hidden dependency.

What is your TODO about, is it that you might be able read the list of the shape parameters directly from the ONNX file?

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -0,0 +1,352 @@
...
+    // pass data through model
+    try
+    {
+      std::vector<Ort::Value> output_tensors = m_ortSession->Run(m_input_names,
+                                                                 input_tensors,
+                                                                 m_output_names,
+                                                                 Ort::RunOptions{nullptr});
+
+      // double-check the dimensions of the output tensors
+      // NOTE: the number of output tensors is equal to the number of output nodes specifed in the Run() call
```
kjvbrt, line 331: specifed -> specified

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -0,0 +1,352 @@
...
+	  int position = std::distance(shapeParameters.begin(), it);
+	  m_inputPositionsInShapeParameters.push_back(position);
+	  info() << "Feature " << feature << " found in position " << position << " of shapeParameters" << endmsg;
+	}
+      else {
+	m_inputPositionsInShapeParameters.push_back(-1);
+      }
+    }
+    else {
+      // for the other features, check if they are in the shape paramaters
```
kjvbrt, line 75: paramaters -> parameters

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: yes, the TODO was about that, but it seems not possible unfortunately

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: fixed

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: fixed

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: changed to vector of short int

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
giovannimarchiori: Hi @kjvbrt I implemented the parsing of the json with nlohmann. The json stores all relevant info from the training of the photonID algorithm and is generated automatically by the tool that trains the BDT and creates the onnx model, so using it rather than copying to the k4run steering file in my opinion provides some better consistency.
I saw that you merged some PRs so I will probably have to rebase my PR. Apart from that is there anything else I should change? Thanks a lot!

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -0,0 +1,361 @@
...
+  // Open the JSON file
+  std::ifstream file(mvaInputsFileName);
+  if (!file.is_open()) {
+    error() << "Error opening file: " << mvaInputsFileName << endmsg;
+    return StatusCode::FAILURE;
+  }
+
+  // Parse the JSON file
+  json j;
+  file >> j;
```
kjvbrt, line 176: What happens if the file contains non valid JSON?

giovannimarchiori: failing miserably I guess, need to probably put this in some try .. catch block. On my todo for next week

giovannimarchiori: fixed

giovannimarchiori:
Hi @kjvbrt , I implemented the handling of exceptions in reading the JSON and rebased - is there anything left for me to do?

Thx, Giovanni

kjvbrt:
Hi @giovannimarchiori, nice :)
Merging...

PR #88, opened 2024-07-01
Add support for reconstruction of "turbine" endcap ecal

varnes:
This PR is a first pass at including the "turbine" ecal endcap into reconstruction.  At the moment only the sliding windows algorithm for creating clusters is supported.  In addition to adding code to support this geometry, there was also a change in CaloTowerToolFCCee::CellsIntoTowers, where segmentation->theta(cell.getCellID()) was changed to cellTheta (which was already calculated from the cell position).  This change means that the segmentation is no longer required by the CellsIntoTowers method, so it was removed from the list of passed parameters.

Further details about the turbine geometry and the segmentation used for its reconstruction can be found at   https://indico.cern.ch/event/1298458/contributions/5977217/attachments/2875296/5035181/FCC_week_June_2024.pdf

This PR is intended to be in concert with a similar PR for k4geo.


kjvbrt:
Hi @varnes, can you add link to the k4geo PR?

varnes:
Hi @kjvbrt ,

  The corresponding k4geo PR is https://github.com/key4hep/k4geo/pull/347.

Best regards,
Erich

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -124,7 +125,6 @@ class CaloTowerToolFCCee : public GaudiTool, virtual public ITowerToolThetaModul
    *   @param[in] aSegmentation Segmentation of the calorimeter
    */
   void CellsIntoTowers(std::vector<std::vector<float>>& aTowers, const edm4hep::CalorimeterHitCollection* aCells,
-                       dd4hep::DDSegmentation::Segmentation* aSegmentation, SegmentationType aType,
```
giovannimarchiori, line 127: if you change the signature of this function, please update also the documentation just above (remove   @param[in] aSegmentation Segmentation of the calorimeter, also please add the bool param)

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -289,6 +296,7 @@ uint CaloTowerToolFCCee::idTheta(float aTheta) const {
 // aPhi Position of the calorimeter cell in phi
 uint CaloTowerToolFCCee::idPhi(float aPhi) const {
   uint id = floor((aPhi + m_phiMax) / m_deltaPhiTower);
+  // uint id = floor(aPhi / m_deltaPhiTower);
```
giovannimarchiori, line 299: remove this commented line (confusing)
By the way, note for self and @BrieucF - I am puzzled by the way we index the towers... why not  floor((aPhi - m_phiMin) / m_deltaPhiTower);, so that the index will go from 0 to m_nPhiTowers-1? 
With this calculation, the index goes from (phiMin+phiMax)/deltaPhi to (phiMin+phiMax)/deltaPhi + m_nPhiTowers-1, and this is what seems to be assumed by the phiNeighbour method, and the current code only gives the same result by accident (because phiMin=-pi, phiMax=pi and thus phiMin+phiMax = 0 and 2 phiMax = phiMax - phiMin = 2pi)

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -184,11 +194,9 @@ void CaloTowerToolFCCee::towersNumber(int& nTheta, int& nPhi) {
   tmpPair = retrievePhiThetaExtrema(m_hcalFwdSegmentation, m_hcalFwdSegmentationType);
   listPhiMax.push_back(tmpPair.first);
   listThetaMax.push_back(tmpPair.second);
-
   // Maximum theta & phi of the calorimeter system
   m_phiMax = *std::max_element(listPhiMax.begin(), listPhiMax.end());
   m_thetaMax = *std::max_element(listThetaMax.begin(), listThetaMax.end());
-  debug() << "Detector limits: phiMax " << m_phiMax << " thetaMax " << m_thetaMax << endmsg;
```
giovannimarchiori, line 191: why removing this debug statement? it doesn't hurt..

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -357,15 +349,16 @@ void CaloTowerToolFCCee::CellsIntoTowers(std::vector<std::vector<float>>& aTower
       float cellPhi   = atan2(cellY, cellX);
       iTheta = idTheta(cellTheta);
       iPhi = idPhi(cellPhi);
-      if (aType == SegmentationType::kPhiTheta) {
-        aTowers[iTheta][phiNeighbour(iPhi)] += cell.getEnergy() * sin(segmentationHCal->theta(cell.getCellID()));
-      } else {
-        aTowers[iTheta][phiNeighbour(iPhi)] += cell.getEnergy() * sin(segmentation->theta(cell.getCellID()));
-      }
+      aTowers[iTheta][phiNeighbour(iPhi)] +=
+	//        cell.getEnergy() * sin(segmentation->theta(cell.getCellID()));
```
giovannimarchiori, line 353: remove the commented line. 
The change assumes that cells are already positioned and we get theta directly from the cell positions rather than their cellID, but this hypothesis was anyway already used in other parts of the code before (to calculate iTheta and iPhi from cellTheta and cellPhi) so I think we can go ahead with this change and simplify the code by removing the need of passing the segmentation class to the cellsInTowers method

giovannimarchiori (commented):
Hi @varnes ,

thanks a lot for this PR!

Overall it looks good to me, but I think that now it is quite evident from the code that we don't need anymore the tool to have all the members of type dd4hep::DDSegmentation::Segmentation* and of type SegmentationType, because the only place where the segmentations are used is in the function that retrieves the phi and theta extremes of the readout in CaloTowerToolFCCee::retrievePhiThetaExtrema.
I think you can remove all those members, remove the code that retrieves the segmentation in initialize(), and modify retrievePhiThetaExtrema to accept the name of the readout rather than segmentation pointer + type. You can retrieve (temporarily) the segmentation in retrievePhiThetaExtrema.
And basically all code of towersNumbers except the last two lines could go into initialize() to calculate at the beginning of the alg the max theta and phi

If that is clear and you want to do it in this PR, that'd be great for me, otherwise we can postpone it to a separate PR. 
Pinging also @BrieucF

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
varnes: Commented line removed

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
varnes: Commented line added back in

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
varnes: Commented line removed

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
varnes: Changed as suggested

varnes:
Hi @giovannimarchiori ,

 I'm happy in principle with removing the members that are no longer needed, and making the related changes you suggest.  However between ICHEP and some vacation, I will not have many free cycles between now and 1 August.  

-- Erich

varnes:
Hi @giovannimarchiori ,

  The segmentation members have been removed, and the other changes you suggested are now implemented.  There does still need to be a check that the segmentations are read in correctly (and therefore that phimax and thetamax are set correctly), so I added a set of bool members to keep track of that.

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -110,24 +117,39 @@ StatusCode CaloTowerToolFCCee::finalize() {
   return GaudiTool::finalize();
 }
 
-std::pair<double, double> CaloTowerToolFCCee::retrievePhiThetaExtrema(dd4hep::DDSegmentation::Segmentation* aSegmentation, SegmentationType aType) {
+std::pair<double, double> CaloTowerToolFCCee::retrievePhiThetaExtrema(Gaudi::Property<std::string> aReadoutName) {
```
giovannimarchiori, line 120: can't the argument simply be a std::string?

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -121,15 +116,17 @@ class CaloTowerToolFCCee : public GaudiTool, virtual public ITowerToolThetaModul
...
+   *   @param[in] fillTowerCells If true, make a list of the cells in each tower
    */
   void CellsIntoTowers(std::vector<std::vector<float>>& aTowers, const edm4hep::CalorimeterHitCollection* aCells,
-                       dd4hep::DDSegmentation::Segmentation* aSegmentation, SegmentationType aType,
                        bool fillTowersCells);
-  /**  Check if the readout name exists. If so, it returns the segmentation.
+  /**  Find the maximum phi, theta covered by a readout
+   *   @param[in] aReadoutName Readout name to be checked for maximum phi, theta
+   */
+  std::pair<double, double> retrievePhiThetaExtrema(Gaudi::Property<std::string> aReadoutName);
```
giovannimarchiori, line 126: can't the argument simply be a std::string?

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -37,69 +43,70 @@ StatusCode CaloTowerToolFCCee::initialize() {
     return StatusCode::FAILURE;
   }
 
-  // check if readouts exist & retrieve Module-Theta segmentations
-  // if readout does not exist, reconstruction without this calorimeter part will be performed
-  std::pair<dd4hep::DDSegmentation::Segmentation*, SegmentationType> tmpPair;
+    std::vector<double> listPhiMax;
```
giovannimarchiori, line 46: please fix indentation

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -37,69 +43,70 @@ StatusCode CaloTowerToolFCCee::initialize() {
...
+  std::vector<double> listThetaMax;
+  listPhiMax.reserve(7);
+  listThetaMax.reserve(7);
```
giovannimarchiori, line 49: note for myself: we might want to refactor this tool in a later PR so that it accepts a list of cell collections and a list of readout names, making it more agnostic of the number of calorimeter sub detectors... 
Then this hardcoded 7 numbers would be the length of the list

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -37,69 +43,70 @@ StatusCode CaloTowerToolFCCee::initialize() {
...
-  m_ecalBarrelSegmentation = tmpPair.first;
-  m_ecalBarrelSegmentationType = tmpPair.second;
-  if (tmpPair.first != nullptr && tmpPair.second == SegmentationType::kWrong) {
-    error() << "Wrong type of segmentation" << endmsg;
-    return StatusCode::FAILURE;
-  }
-  if (m_useHalfTower) {
-    m_decoder = m_geoSvc->getDetector()->readout(m_ecalBarrelReadoutName).idSpec().decoder();
+  std::pair<double, double> tmpPair;
+  tmpPair = retrievePhiThetaExtrema(m_ecalBarrelReadoutName);
```
giovannimarchiori, line 52: I would remove all the m_XXXSegmentationOK booleans and make the job fail with a return StatusCode::FAILURE; If the segmentation is not OK.
We can also do it in a follow-up PR

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -215,63 +205,62 @@ uint CaloTowerToolFCCee::buildTowers(std::vector<std::vector<float>>& aTowers, b
   const edm4hep::CalorimeterHitCollection* ecalBarrelCells = m_ecalBarrelCells.get();
   debug() << "Input Ecal barrel cell collection size: " << ecalBarrelCells->size() << endmsg;
   // Loop over a collection of calorimeter cells and build calo towers
-  if (m_ecalBarrelSegmentation != nullptr) {
-    CellsIntoTowers(aTowers, ecalBarrelCells, m_ecalBarrelSegmentation, m_ecalBarrelSegmentationType, fillTowersCells);
+  if (m_ecalBarrelSegmentationOK) {
```
giovannimarchiori, line 208: if we remove the m_XXXSegmentationOK booleans, here we could check instead if the cell collection name was not an empty string. 
In a later PR, if we refactor the code to pass lists of cell collections and readouts, we won't even need to do this check - we would just loop over the list of cell collection names, retrieve the cells and call CellsIntoTowers for each cell collection

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -198,6 +181,15 @@ class CaloTowerToolFCCee : public GaudiTool, virtual public ITowerToolThetaModul
...
   /// decoder: only for barrel
   dd4hep::DDSegmentation::BitFieldCoder* m_decoder;
+  // Set of bools to record whether a proper segmentation was found
+  bool m_ecalBarrelSegmentationOK;
+  bool m_ecalEndcapSegmentationOK;
+  bool m_ecalFwdSegmentationOK;
+  bool m_hcalBarrelSegmentationOK;
+  bool m_hcalExtBarrelSegmentationOK;
+  bool m_hcalEndcapSegmentationOK;
+  bool m_hcalFwdSegmentationOK;
```
giovannimarchiori, line 191: I think we can get rid of these, but can also be done in a separate PR later

giovannimarchiori (commented):
HI @varnes 
thanks a lot for these updates.
There are 2-3 little things to be fixed (one indentation and passing directly a std::string to the CaloTowerToolFCCee::retrievePhiThetaExtrema method). We could also get rid of the booleans m_xxxSegmentationOK and return a StatusCode::Failure when the segmentation is not OK, but we can do it also in a separate PR if you don't have time. I also think I'll have a further shot at this code in a later PR to make it agnostic about the number of calorimeter sub-detectors, so the boolean removal could also go in there.

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
varnes: The bool class members have now been removed, and the interface to retrievePhiThetaExtrema changed so that it returns StatusCode.  The pair to hold the phi and theta max values is now passed by reference.

giovannimarchiori (approved):
Hi @varnes thanks a lot, the PR looks good to me now

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -182,6 +182,17 @@ class CaloTowerToolFCCee : public AlgTool, virtual public ITowerToolThetaModule
...
+<<<<<<< HEAD
+=======
+  // Set of bools to record whether a proper segmentation was found
+  bool m_ecalBarrelSegmentationOK;
+  bool m_ecalEndcapSegmentationOK;
+  bool m_ecalFwdSegmentationOK;
+  bool m_hcalBarrelSegmentationOK;
+  bool m_hcalExtBarrelSegmentationOK;
+  bool m_hcalEndcapSegmentationOK;
+  bool m_hcalFwdSegmentationOK;
```
giovannimarchiori, line 194: there was a problem with rebasing it seems, as some old code that was removed came back...

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -182,6 +182,17 @@ class CaloTowerToolFCCee : public AlgTool, virtual public ITowerToolThetaModule
...
+>>>>>>> 0bdbac8 (Remove segmentation members)
```
giovannimarchiori, line 195: to be removed

giovannimarchiori:
Hi @varnes, we merged a PR to fix some tests that were failing due to the EventCounter, could you please rebase your branch (and fix conflicts in case)? Hopefully then all tests will pass and we can merge it. 
Thanks!

varnes:
Hi @giovannimarchiori ,

 I think it should be OK now...

giovannimarchiori:
Hi @varnes ,

as you can see from the error messages above, the tests for the nightly builds are failing. If you click on Details you will see that there are 3 tests failing:

- FCCeeLAr_simRecoAllegroV3: 
```
File "RecFCCeeCalorimeter/tests/options/run_thetamodulemerged.py", line 503, in <module>
    fileName=os.environ['FCCBASEDIR'] +
  File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-07-24/x86_64-almalinux9-gcc11.4.1-opt/python/3.10.14-a23y4f/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'FCCBASEDIR'
```

=> if you want topoclustering to be run in this test, can you please use the following files
    neighboursMap = "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/neighbours_map_ecalB_thetamodulemerged.root"
    noiseMap = "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/cellNoise_map_electronicsNoiseLevel_ecalB_thetamodulemerged.root"

so that FCCBASEDIR is not needed?
 
- FCCeeLAr_benchmarkCorrection:
```
File "RecFCCeeCalorimeter/tests/options/run_thetamodulemerged_SW_benchmarkCorr.py", line 402
    cellPositionEcalEndcapTool = CellPositionsECalEndcapTurbineSegTool(
IndentationError: unexpected indent
```
=> there is a spurious space at the beginning of line 402 in this file, can you please fix it?

-  FCCeeLAr_runxtalk .
```
*** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================

#5  <signal handler called>
#6  0x00007f85161f8935 in CaloTowerToolFCCee::CellsIntoTowers(std::vector<std::vector<float, std::allocator<float> >, std::allocator<std::vector<float, std::allocator<float> > > >&, edm4hep::CalorimeterHitCollection const*, bool) () from /k4reccalorimeter/build/RecFCCeeCalorimeter/libk4RecFCCeeCalorimeterPlugins.so
```

I think I had fixed this error in the code I sent you by mail. I have the impression that you put the file runEcalBarrel_ReconstructionTopoClusters_crosstalk.py in the wrong directory - you added it as a new test to RecFCCeeCalorimeter/tests/options/ while it should have been in RecCalorimeter/tests/options/

Please make these 3 changes and push, and check if the tests in the nightlies pass successfully as they should.

@BrieucF : if there are still errors after the next attempt, it might be more efficient if we merge the PR and I fix the tests in a separate PR.

varnes:
Hi @giovannimarchiori ,

 Would it be best for me to just update these scripts to the versions that are currently in the main branch, or should I take the ones from your email? 

Thanks,
Erich

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
giovannimarchiori: I still see them in the latest commit, will merge and fix (with the tests) in a new PR

giovannimarchiori:
Hi @varnes ,

it was quicker for me to merge your PR and fix the remaining issues in https://github.com/HEP-FCC/k4RecCalorimeter/pull/106

Thanks, cheers,
   Giovanni

PR #90, opened 2024-07-04
Photon ID shape parameter: correct very small negative values caused by computational precision in width of theta/module

dasphy:
Follow-up of #79 . In the width calculation:
```double _w_theta_3Bin = sqrt(theta2_E_3Bin / sum_E_3Bin - std::pow(theta_E_3Bin / sum_E_3Bin, 2));```
Mathematically the value inside sqrt() should never be negative, but a very small negative value might be produced, due to the precision of calculation in code.

A threshold ```-1e-9``` is set to correct this:
```if (_w_theta_3Bin2 < 0. && _w_theta_3Bin2 > -1e-9)    _w_theta_3Bin2 = 0. ;```

Tagging @giovannimarchiori , @kjvbrt 

Cheers, Tong

### CODE REVIEW

**Comment by dasphy** [comment]
Yes I think fabs() is better. I just pushed changes.

**Comment by giovannimarchiori** [comment]
@kjvbrt I don't like too much this solution, without even adding a comment to the code to explain the fabs, nor a warning in the output. I would have preferred the option that if the argument of sqrt is slightly negative the width is set to zero and a warning message is print to log, giving also the value of wtheta2.
With fabs we're trusting the fact that wtheta2 is negative only because of small rounding issues and not because of some buggy calculation upstream.

**Comment by dasphy** [comment]
I added warnings in the log and set the value to 0 if ```w_theta2``` is slightly negative. please have a look at the changes.
Cheers, Tong

File: RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp [hunk 1]
```diff
@@ -633,16 +633,31 @@ StatusCode AugmentClustersFCCee::execute([[maybe_unused]] const EventContext &ev
...
-	  double w_theta;
           if (m_do_widthTheta_logE_weights) {
-            w_theta = (sumWeightLayer[layer+startPositionToFill] != 0.) ? sqrt(theta2_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill], 2)) : 0. ;
+            double w_theta2 = theta2_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill], 2);
+            // Correct very small but negative values caused by computational precision
+            if (w_theta2 < 0. && w_theta2 > -1e-9) {
+              warning() << "w_theta2 in theta width calculation is negative: " << w_theta2 << ", setting it to 0." << endmsg;
+              w_theta2 = 0. ;
+            }
+            width_theta[layer+startPositionToFill] = (sumWeightLayer[layer+startPositionToFill] > 0.) ? std::sqrt(w_theta2) : 0. ;
```
kjvbrt, line 643:
```suggestion
            double w_theta2 = theta2_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill] - std::pow(theta_E_layer[layer+startPositionToFill] / sumWeightLayer[layer+startPositionToFill], 2);
            // Very small but negative values caused by computational precision are allowed,
            // otherwise crash.
            if (w_theta2 < -1e-9) {
              error() << "w_theta2 in theta width calculation is negative: " << w_theta2 << endmsg;
              return StatusCode::FAILURE;
            }
            width_theta[layer+startPositionToFill] = (sumWeightLayer[layer+startPositionToFill] > 0.) ? std::sqrt(std::fabs(w_theta2)) : 0. ;
```

kjvbrt:
Hi, what about having a hard crash or warning if the variable is negative more that `-1e-9` and otherwise use `std::fabs`?

The test now notifies only when the negative value is very small.

giovannimarchiori:
> Hi, what about having a hard crash or warning if the variable is negative more that `-1e-9` and otherwise use `std::fabs`?
> 
> The test now notifies only when the negative value is very small.

Seems a good idea to me

kjvbrt:
Can you do the same change for all the cases?

dasphy:
Sounds good to me, let me modify the code accordingly

kjvbrt:
Nice, looks good to me. Will merge shortly...

PR #91, opened 2024-07-04
update CellPositionsTool for HCal Allegro

mmlynari:
- added new CellPositionsHCalPhiThetaSegTool to calculate radii of HCal barrel and HCal endcap on the flight from the geometry. Xml files that include extensions with radii of each layer and cells dimensions are HCalBarrel_TileCal_v02.xml and HCalEndcaps_ThreeParts_TileCal_v02.xml (now included in ALLEGRO_o1_v03.xml) 

- this should be merged together with [PR#350](https://github.com/key4hep/k4geo/pull/350) in k4geo

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
+++ b/RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
@@ -33,11 +33,10 @@ StatusCode CellPositionsHCalPhiThetaSegTool::initialize()
   if (m_segmentation == nullptr)
   {
     error() << "There is no phi-theta segmentation!!!!" << endmsg;
-    // return StatusCode::FAILURE;
+    return StatusCode::FAILURE;
   }
   // Take readout bitfield decoder from GeoSvc
   m_decoder = m_geoSvc->getDetector()->readout(m_readoutName).idSpec().decoder();
-  m_volman = m_geoSvc->getDetector()->volumeManager();
 
   // check if decoder contains "layer"
   std::vector<std::string> fields;
@@ -49,9 +48,10 @@ StatusCode CellPositionsHCalPhiThetaSegTool::initialize()
   if (iter == fields.end())
   {
     error() << "Readout does not contain field: 'layer'" << endmsg;
+    return StatusCode::FAILURE;
   }
 
-  // needed to retrieve radii from the dd4hep extension
+  // retrieve radii from the LayeredCalorimeterData extension
   dd4hep::Detector* detector = m_geoSvc->getDetector();
   if (!detector)
   {
@@ -73,8 +73,7 @@ StatusCode CellPositionsHCalPhiThetaSegTool::initialize()
     return StatusCode::FAILURE;
   }
 
-  // Debug information to verify the layers retrieved
-  // m_layersRetrieved is a pointer, then it needs to point to the address of the vector
+  // Debug information to check the number of layers retrieved from the LayeredCalorimeterData extension
   m_layersRetrieved = &(theExtension->layers);
   debug() << "Number of layers retrieved: " << m_layersRetrieved->size() << endmsg;
 
@@ -93,6 +92,14 @@ StatusCode CellPositionsHCalPhiThetaSegTool::initialize()
 
   if (m_detectorName=="HCalThreePartsEndcap")
   {
+    // Check that the vector containing number of layers in each cylinder is provided
+    if (m_numLayersHCalThreeParts.empty())
+    {
+       error() << "The vector m_numLayersHCalThreeParts is empty." << endmsg;
+       return StatusCode::FAILURE;
+    }
+    // Check that the total number of layers provided in the steering file
+    // matches the total number of layers in the geometry xml file
     for (unsigned int i=0; i<m_numLayersHCalThreeParts.size(); i++) 
     {
       numLayersProvided += m_numLayersHCalThreeParts[i];
@@ -164,12 +171,16 @@ int CellPositionsHCalPhiThetaSegTool::layerId(const uint64_t &aCellId)
   return layer;
 }
 
+// calculate layer radii from LayeredCalorimeterData extension
+// which is included in the geometry description
 std::vector<double> CellPositionsHCalPhiThetaSegTool::calculateLayerRadii(unsigned int startIndex, unsigned int endIndex) {
   std::vector<double> radii;
 
   for (unsigned int idxLayer = startIndex; idxLayer < endIndex; ++idxLayer) {
     const dd4hep::rec::LayeredCalorimeterStruct::Layer & theLayer = m_layersRetrieved->at(idxLayer);
+    // inner radius of a given layer
     double layerInnerRadius = theLayer.distance;
+    // radial dimension of a given layer
     double layerThickness = theLayer.sensitive_thickness;
 
     // Calculate mid-radius for the current layer (layerInnerRadius+layerOuterRadius)/2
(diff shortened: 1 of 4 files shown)


mmlynari:
Tagging @faltovaj , @giovannimarchiori  and @BrieucF  to have a look at the proposed changes

giovannimarchiori:
Hi @mmlynari ,
thanks for this MR!
Have you tested that the code works as expected, and positions the cells properly in both the barrel and endcap calorimeters?
Also, the new version of the tool seems to be used both by barrel and endcap, so it might be good to rename it and remove "Barrel" from the name?
In the endcap, the radii are still dimensions along rho (is the endcap composed of nested disks) or do they correspond to some longitudinal granularity (I will also check the c++ that builds the detector to have a clearer picture)?

mmlynari:
Hello @giovannimarchiori , 
yes, I have tested that the code correctly calculates radii of each layer in both barrel and endcap. I still need to have a closer look at the calculation of the z-coordinate in the endcap (at the first glance it seems to be off in a few cases), but this might point to a problem in the geometry itself. So I thought that we might have this first version of the cell positions tool with automated calculation of the radius and if more updates are needed (especially for the endcap), I will add them when I have better understanding of what is happening. 
Right, I will remove 'Barrel' from the name. Should we still keep the original CellPositionsHCalBarrelPhiThetaSegTool and also have this new CellPositionsHCalPhiThetaSegTool? 
In the endcap, the layer radii are in dimensions along rho. The geometry is very similar to the barrel, but in the endcap we have three cylinders along z, each with different inner radius and different layer number/dimensions.

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
```diff
@@ -0,0 +1,211 @@
...
+    error() << "Unable to locate Geometry service." << endmsg;
+    return StatusCode::FAILURE;
+  }
+  // get PhiTheta segmentation
+  m_segmentation = dynamic_cast<dd4hep::DDSegmentation::FCCSWGridPhiTheta_k4geo *>(
+      m_geoSvc->getDetector()->readout(m_readoutName).segmentation().segmentation());
+  if (m_segmentation == nullptr)
+  {
+    error() << "There is no phi-theta segmentation!!!!" << endmsg;
+    // return StatusCode::FAILURE;
```
giovannimarchiori, line 36: why there's an error message here not followed by a return StatusCode::FAILURE? Did you forget uncommenting the line or there is a good reason? In the latter case, you should not use an error(), rather a warning(), and handle the else accordingly

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
```diff
@@ -0,0 +1,211 @@
...
+  // check if decoder contains "layer"
+  std::vector<std::string> fields;
+  for (uint itField = 0; itField < m_decoder->size(); itField++)
+  {
+    fields.push_back((*m_decoder)[itField].name());
+  }
+  auto iter = std::find(fields.begin(), fields.end(), "layer");
+  if (iter == fields.end())
+  {
+    error() << "Readout does not contain field: 'layer'" << endmsg;
```
giovannimarchiori, line 51: should we return a StatusCode::FAILURE here? or is it legitimate that the code continues running if the readout does not contain the layer field?

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
```diff
@@ -0,0 +1,72 @@
...
+class IGeoSvc;
+namespace DD4hep {
+namespace DDSegmentation {
+class Segmentation;
+}
+}
+
+/** @class CellPositionsHCalPhiThetaSegTool
+ *
+ */
```
giovannimarchiori, line 35: the documentation could be expanded a bit (adding also some doxygen style documentation for the methods below)

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
```diff
@@ -0,0 +1,72 @@
...
+  SmartIF<IGeoSvc> m_geoSvc;
+  /// Name of the calorimeter readout
+  Gaudi::Property<std::string> m_readoutName{this, "readoutName", "HCalBarrelReadout"};
+  /// Name of the calorimeter 
+  Gaudi::Property<std::string> m_detectorName{this, "detectorName", "HCalBarrel"};
+
+  std::vector<double> m_radii;
+  dd4hep::DDSegmentation::FCCSWGridPhiTheta_k4geo* m_segmentation;
+  dd4hep::DDSegmentation::BitFieldCoder* m_decoder;
+  dd4hep::VolumeManager m_volman;
```
giovannimarchiori, line 67: I think this is not used by your code and can be removed

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
```diff
@@ -0,0 +1,211 @@
...
+  m_segmentation = dynamic_cast<dd4hep::DDSegmentation::FCCSWGridPhiTheta_k4geo *>(
+      m_geoSvc->getDetector()->readout(m_readoutName).segmentation().segmentation());
+  if (m_segmentation == nullptr)
+  {
+    error() << "There is no phi-theta segmentation!!!!" << endmsg;
+    // return StatusCode::FAILURE;
+  }
+  // Take readout bitfield decoder from GeoSvc
+  m_decoder = m_geoSvc->getDetector()->readout(m_readoutName).idSpec().decoder();
+  m_volman = m_geoSvc->getDetector()->volumeManager();
```
giovannimarchiori, line 40: not used?

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
```diff
@@ -0,0 +1,72 @@
...
+  // layer radii calculated on the flight from the geometry 
+  const std::vector<dd4hep::rec::LayeredCalorimeterStruct::Layer>* m_layersRetrieved;
```
kjvbrt, line 69: Can you initialize this to `nullptr`?

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
mmlynari: this was not intentional, just a copy-paste from the CellPositionsHCalBarrelPhiThetaSegTool. In the new version of the code, it will return status code failure.

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
mmlynari: added StatusCode::FAILURE since the layer info is needed for the position calculations

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
mmlynari: done

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
mmlynari: right, removed m_volman

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.cpp
mmlynari: m_volman is not used, so it was removed

RecFCCeeCalorimeter/src/components/CellPositionsHCalPhiThetaSegTool.h
mmlynari: sure, done

giovannimarchiori:
Hi @mmlynari,
now that the k4geo PR has been merged, we are close to merging also this PR, but you need to update the code (don't worry, the changes are minimal) following the migration from GaudiTool to AlgTool, as you can see here: https://github.com/HEP-FCC/k4RecCalorimeter/pull/101/files#diff-00d25010e6ca61c5b6c8a8f09cfc3af003cb599649fc729de7c2c30353693037

Basically replace everywhere GaudiTool with AlgTool in .h and .cpp, and 
#include "GaudiAlg/GaudiTool.h"
with 
#include "GaudiKernel/AlgTool.h"
in cpp

Then push to the branch used for the PR.
If things are OK, the tests for the nightlies should be succeeding and then we can merge

Thanks a lot!

giovannimarchiori (approved):

PR #92, opened 2024-07-09
[CreateCaloCells] Switch cross-talk to false by default

BrieucF:


BrieucF:
I merge this now to move forward with test fixing (e.g. in FCC-config), @zwu0922 will check that the x-talk tool still behaves as expected and we will fix things in another PR if needed.

PR #95, opened 2024-07-09
Adding more improvements to the jet tool

jroloff:
I've made a few changes to address some of the remaining comments in !80, including using a MultiTransformer for the truth jet tool, and making sure that the truth jet test produced jets.


kjvbrt:
nice :)

PR #98, opened 2024-07-19
Consolidate tools handling noise

giovannimarchiori:
Consolidate code of tools handling noise, improve readability, remove duplications, add tool to add noise to cells for fcc-ee.
Needs https://github.com/key4hep/k4FWCore/pull/214


RecCalorimeter/src/components/ConstNoiseTool.h
```diff
@@ -34,34 +37,33 @@ class ConstNoiseTool : public GaudiTool, virtual public INoiseConstTool {
...
+  double getNoiseRMSPerCell(uint64_t aCellID);
   double getNoiseOffsetPerCell(uint64_t aCellID);
 
 private:
-  std::map<uint,double> m_systemNoiseConstMap;
+  std::map<uint,double> m_systemNoiseRMSMap;
   std::map<uint,double> m_systemNoiseOffsetMap;
 
-  /// Cell thresholds (1Sigma) in GeV:
+  /// List of subdetector names (I think it is better to pass the names and retrieve the IDs in the C++)
```
BrieucF, line 47: On drawback of this approach is that it will not work for a detector that passes the system id directly as a number in the xml. But probably no detector like this will use the tool. 

If we keep this approach, can you add as comment here that the name of the detector *must* match what is defined in DectDimentsion?

RecCalorimeter/src/components/NoiseCaloCellsFlatTool.cpp
```diff
@@ -27,19 +27,20 @@ StatusCode NoiseCaloCellsFlatTool::initialize() {
...
 void NoiseCaloCellsFlatTool::addRandomCellNoise(std::unordered_map<uint64_t, double>& aCells) {
   std::for_each(aCells.begin(), aCells.end(),
-                [this](std::pair<const uint64_t, double>& p) { p.second += (m_gauss.shoot() * m_cellNoise); });
+                [this](std::pair<const uint64_t, double>& p) { p.second += (m_cellNoiseOffset + (m_gauss.shoot() * m_cellNoiseRMS)); });
 }
 
 void NoiseCaloCellsFlatTool::filterCellNoise(std::unordered_map<uint64_t, double>& aCells) {
   // Erase a cell if it has energy below a threshold
-  double threshold = m_filterThreshold * m_cellNoise;
+  double threshold = m_cellNoiseOffset + m_filterThreshold * m_cellNoiseRMS;
```
BrieucF, line 43: nice catch, thanks

RecFCCeeCalorimeter/src/components/NoiseCaloCellsVsThetaFromFileTool.h
```diff
@@ -0,0 +1,127 @@
...
+
+// Root
+class TH1F;
+
+/** @class NoiseCaloCellsVsThetaFromFileTool
+ *
+ *  Tool for calorimeter noise
+ *  Access noise constants from TH1F histogram (noise vs. theta)
+ *  createRandomCellNoise: Create random CaloHits (gaussian distribution) for the vector of cells
+ *  filterCellNoise: remove cells with energy below threshold*sigma from the vector of cells
```
BrieucF, line 27: Won't we include this filtering in CreateCaloCells?

BrieucF (commented):
Looks good to me!

BrieucF:
The tests show: https://github.com/HEP-FCC/k4RecCalorimeter/actions/runs/10079966073/job/27868570990?pr=98#step:3:264

giovannimarchiori:
> The tests show: https://github.com/HEP-FCC/k4RecCalorimeter/actions/runs/10079966073/job/27868570990?pr=98#step:3:264

The tests fail because they need this: https://github.com/key4hep/k4FWCore/pull/214

RecFCCeeCalorimeter/src/components/NoiseCaloCellsVsThetaFromFileTool.h
giovannimarchiori: the method is indeed called in CreateCaloCells: https://github.com/HEP-FCC/k4RecCalorimeter/blob/a902075091a928d40c3b6290792d1e2296401190/RecCalorimeter/src/components/CreateCaloCells.cpp#L108

What would you like to change?

BrieucF: It can be done in another PR, but the filtering of cells could be isolated away from the noise tool. One may want to filter cells even without adding noise.

RecCalorimeter/src/components/ReadNoiseFromFileTool.cpp
```diff
@@ -234,12 +233,12 @@ double ReadNoiseFromFileTool::getNoiseOffsetPerCell(uint64_t aCellId) {
...
-  double totalNoise = sqrt(pow(elecNoise, 2) + pow(pileupNoise, 2)) * m_scaleFactor;
+  double totalNoiseOffset = sqrt(elecNoiseOffset*elecNoiseOffset + pileupNoiseOffset*pileupNoiseOffset) * m_scaleFactor; // shouldnt the offset be summed linearly?
 
-  if (totalNoise < 1e-6) {
-    warning() << "Zero noise: cell eta " << cellEta << " layer " << cellLayer << " noise " << totalNoise << endmsg;
-  }
+  // No warning: offset is usually zero or close
+  //if (totalNoiseOffset < 1e-6) {
+  //  warning() << "Zero noise offset: cell eta " << cellEta << " layer " << cellLayer << " noise " << totalNoiseOffset << endmsg;
+  //}
```
kjvbrt, line 241: Can you remove this comment out part? And put comment about the offset value to line `236`?

RecCalorimeter/src/components/TopoCaloNoisyCells.cpp
```diff
@@ -44,7 +44,7 @@ StatusCode TopoCaloNoisyCells::initialize() {
   double readNoisyCells;
   double readNoisyCellsOffset;
   tree->SetBranchAddress("cellId", &readCellId);
-  tree->SetBranchAddress("noiseLevel", &readNoisyCells);
+  tree->SetBranchAddress("noiseLevel", &readNoisyCells); // would be better to call branch noiseRMS rather than noiseLevel
```
kjvbrt, line 47: What is blocking to do it?

RecFCCeeCalorimeter/src/components/NoiseCaloCellsVsThetaFromFileTool.cpp
```diff
@@ -0,0 +1,276 @@
...
+  double elecNoiseOffset = 0.;
+  double pileupNoiseOffset = 0.;
+
+  // Get cell coordinates: theta and radial layer
+  double cellTheta = m_cellPositionsTool->xyzPosition(aCellId).Theta();
+  unsigned cellLayer = m_decoder->get(aCellId, m_activeFieldName);
+
+  // All histograms have same binning, all bins with same size
+  // Using the histogram in the first layer to get the bin size
+  unsigned index = 0;
```
kjvbrt, line 246: You can put `index` inside of the if

RecCalorimeter/src/components/TopoCaloNoisyCells.cpp
giovannimarchiori: I am not sure, this would mean that the noise map files have to be regenerated changing name to the rms branch. I would be ok with that but don’t want to disrupt the user who have already noise files at hand. What do you and @BrieucF prefer?

BrieucF: Yes, we can handle this in another PR

PR #100, opened 2024-07-24
WIP: expanded version of CreateCaloCells that also add proper cell positions

giovannimarchiori:
The position added to the cells by CreateCaloCells uses volume ID + segmentation to calculate the position of the cells, but it lacks the z information (maybe the code could have worked but the part that retrieves the z information from the segmentation does not seem to be implemented, and would depend on the segmentation type).

Since we have already tools that translates cellID into position, I wrote a CreatePositionedCaloCells algorithm that works exactly like CreateCaloCells but for the cell position uses the positioning tool to calculate it and save it into the cell.

I think this tool could probably supersede both CreateCaloCells and CreateCaloCellPositionsFCCee, but for the moment I just added a new tool and left there the old ones so that nobody is affected.

Example of usage:

with CreateCaloCells and CreateCaloCellPositionsFCCee:

```
ecalBarrelCellsName = "ECalBarrelCells"
createEcalBarrelCells = CreateCaloCells("CreateECalBarrelCells",
                                         doCellCalibration=True,
                                         calibTool=calibEcalBarrel,
                                         addCellNoise=False,
                                         filterCellNoise=False,
                                         addPosition=True,
                                         OutputLevel=INFO,
                                         hits=ecalBarrelReadoutName,
                                         cells=ecalBarrelCellsName)

cellPositionEcalBarrelTool = CellPositionsECalBarrelModuleThetaSegTool(
    "CellPositionsECalBarrel",
    readoutName=ecalBarrelReadoutName,
    OutputLevel=INFO
)
ecalBarrelPositionedCellsName = ecalBarrelReadoutName + "Positioned"
createEcalBarrelPositionedCells = CreateCaloCellPositionsFCCee(
    "CreateECalBarrelPositionedCells",
    OutputLevel=INFO
)
createEcalBarrelPositionedCells.positionsTool = cellPositionEcalBarrelTool
createEcalBarrelPositionedCells.hits.Path = ecalBarrelCellsName
createEcalBarrelPositionedCells.positionedHits.Path = ecalBarrelPositionedCellsName
```

with CreatePositionedCaloCells:

```
cellPositionEcalBarrelTool = CellPositionsECalBarrelModuleThetaSegTool(
    "CellPositionsECalBarrel",
    readoutName=ecalBarrelReadoutName,
    OutputLevel=INFO
)
ecalBarrelPositionedCellsName = ecalBarrelReadoutName + "Positioned"
createEcalBarrelCells = CreatePositionedCaloCells("CreatePositionedECalBarrelCells",
                                                  doCellCalibration=True,
                                                  positionsTool=cellPositionEcalBarrelTool,
                                                  calibTool=calibEcalBarrel,
                                                  addCellNoise=False,
                                                  filterCellNoise=False,
                                                  OutputLevel=INFO,
                                                  hits=ecalBarrelReadoutName,
                                                  cells=ecalBarrelPositionedCellsName)
```

if this class was promoted to the baseline for creating calo cells, then the cells from digitisation would always be positioned.
Of course it requires a positioning tool translating readout into positions, but this is always needed anyway for further reconstruction steps. The following algorithm could then assume that the input cells are always positioned.

Marking it as draft for the time being until we discuss with @BrieucF  whether this is the way we want to go

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CreatePositionedCaloCells.cpp
+++ b/RecFCCeeCalorimeter/src/components/CreatePositionedCaloCells.cpp
@@ -1,0 +1,197 @@
+#include "CreatePositionedCaloCells.h"
+
+// k4geo
+#include "detectorCommon/DetUtils_k4geo.h"
+
+// k4FWCore
+#include "k4Interface/IGeoSvc.h"
+
+// DD4hep
+//#include "DD4hep/Detector.h"
+//#include "DD4hep/Volumes.h"
+//#include "TGeoManager.h"
+
+// edm4hep
+#include "edm4hep/CalorimeterHit.h"
+
+DECLARE_COMPONENT(CreatePositionedCaloCells)
+
+CreatePositionedCaloCells::CreatePositionedCaloCells(const std::string& name, ISvcLocator* svcLoc) :
+GaudiAlgorithm(name, svcLoc), m_geoSvc("GeoSvc", name) {
+  declareProperty("hits", m_hits, "Hits from which to create cells (input)");
+  declareProperty("cells", m_cells, "The created calorimeter cells (output)");
+
+  declareProperty("positionsTool", m_cellPositionsTool, "Handle for cell positions tool");
+  declareProperty("crosstalkTool", m_crosstalkTool, "Handle for the cell crosstalk tool");
+  declareProperty("calibTool", m_calibTool, "Handle for tool to calibrate Geant4 energy to EM scale tool");
+  declareProperty("noiseTool", m_noiseTool, "Handle for the calorimeter cells noise tool");
+  declareProperty("geometryTool", m_geoTool, "Handle for the geometry tool");
+}
+
+StatusCode CreatePositionedCaloCells::initialize() {
+  StatusCode sc = GaudiAlgorithm::initialize();
+  if (sc.isFailure()) return sc;
+
+  info() << "CreatePositionedCaloCells initialized" << endmsg;
+  info() << "do calibration : " << m_doCellCalibration << endmsg;
+  info() << "add cell noise      : " << m_addCellNoise << endmsg;
+  info() << "remove cells below threshold : " << m_filterCellNoise << endmsg;
+  info() << "emulate crosstalk : " << m_addCrosstalk << endmsg;
+
+  // Initialization of tools
+
+  // Cell position tool
+  if (!m_cellPositionsTool.retrieve()) {
+    error() << "Unable to retrieve the cell positions tool!!!" << endmsg;
+    return StatusCode::FAILURE;
+  }
+  // Cell crosstalk tool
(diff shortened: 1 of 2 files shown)


kjvbrt:
Hi @giovannimarchiori, would this algorithm work also for FCChh?
The question is mostly about the name for the algorithm...

kjvbrt:
Yes

RecFCCeeCalorimeter/src/components/CreatePositionedCaloCells.cpp
```diff
@@ -0,0 +1,192 @@
...
+  declareProperty("geometryTool", m_geoTool, "Handle for the geometry tool");
+}
+
+StatusCode CreatePositionedCaloCells::initialize() {
+  StatusCode sc = Gaudi::Algorithm::initialize();
+  if (sc.isFailure()) return sc;
+
+  info() << "CreatePositionedCaloCells initialized" << endmsg;
+  info() << "do calibration : " << m_doCellCalibration << endmsg;
+  info() << "add cell noise      : " << m_addCellNoise << endmsg;
```
kjvbrt, line 32: Why multiple spaces only here? Either align all lines or none :)

RecCalorimeter/src/components/CreatePositionedCaloCells.cpp
```diff
@@ -0,0 +1,189 @@
...
+    StatusCode sc_prepareCells = m_geoTool->prepareEmptyCells(m_cellsMap);
+    if (sc_prepareCells.isFailure()) {
+      error() << "Unable to create empty cells!" << endmsg;
+      return StatusCode::FAILURE;
+    }
+  }
+
+  // Copy over the CellIDEncoding string from the input collection to the output collection
+  auto hitsEncoding = m_hitsCellIDEncoding.get_optional();
+  if (!hitsEncoding.has_value()) {
```
kjvbrt, line 75: Is this worthy of crashing, can't this be just a warning?

giovannimarchiori: it's exactly as in CreateCaloCells (https://github.com/HEP-FCC/k4RecCalorimeter/blob/02d796b43562ce1c4188d7ef95258a53366a1745/RecCalorimeter/src/components/CreateCaloCells.cpp#L78). Without encoding the input cells can't be properly decoded in terms of layer and theta/eta/phi id which is needed for the various tools (positioning, calibration, ..).
I

giovannimarchiori:
Hi @kjvbrt , I implemented your comments and answered one question. I reran the tests locally and also in GitHub and they ran fine

kjvbrt:
Hi @giovannimarchiori, nice :)
I'm merging this as it draws from the previous work, but I believe that ideally the algorithms like this need to be split into individual tools. And tools converted into algorithms...

PR #101, opened 2024-07-27
Do not use GaudiAlg

jmcarcell:
These are the changes needed not to use GaudiAlg, which will be removed (maybe soon) in Gaudi.


BrieucF:
Looks like something broke

jmcarcell:
Let's merge this after #102 and then check CI again but in principle this is ready and all the tests are passing except the two en #102

BrieucF:
Thanks Juan!

PR #104, opened 2024-07-30
Make sure to properly support newer versions of onnxruntime

tmadlener:
This PR ensures that newer versions of onnxruntime are preferred by CMake and also that newer versions of onnxruntime actually compile (see also https://github.com/HEP-FCC/FCCAnalyses/pull/381 for some more details than here). In short this PR 

- Makes `FindONNXRuntime.cmake` create the an `onnxruntime::onnxruntime` target that works similar (enough) to the one from `onnxruntime`, such that we can simplify the cmake configuration to only use that.
- Makes the c++ code work with older versions (< 1.13) and newer versions using some pre-processor checks.

The CI has been working because it found an ONNXRuntime version 1.10, but I could not build this with a newer version of onnxruntime locally.

- [x] Also includes the changes from #103 so that has to be merged first.

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
+++ b/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
@@ -13,23 +13,18 @@
 #include "edm4hep/Cluster.h"
 #include "edm4hep/ClusterCollection.h"
 #include "edm4hep/CalorimeterHitCollection.h"
+#include <onnxruntime_cxx_api.h>
+
+#include "OnnxruntimeUtilities.h"
 
 DECLARE_COMPONENT(CalibrateCaloClusters)
 
-// convert vector data with given shape into ONNX runtime tensor
-template <typename T>
-Ort::Value vec_to_tensor(std::vector<T> &data, const std::vector<std::int64_t> &shape)
-{
-  Ort::MemoryInfo mem_info =
-      Ort::MemoryInfo::CreateCpu(OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault);
-  auto tensor = Ort::Value::CreateTensor<T>(mem_info, data.data(), data.size(), shape.data(), shape.size());
-  return tensor;
-}
 
 CalibrateCaloClusters::CalibrateCaloClusters(const std::string &name,
                                              ISvcLocator *svcLoc)
     : Gaudi::Algorithm(name, svcLoc),
-      m_geoSvc("GeoSvc", "CalibrateCaloClusters")
+      m_geoSvc("GeoSvc", "CalibrateCaloClusters"),
+      m_ortMemInfo(Ort::MemoryInfo::CreateCpu(OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault))
 {
   declareProperty("inClusters", m_inClusters,
                   "Input cluster collection");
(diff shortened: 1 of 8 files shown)


kjvbrt:
Thanks :)

giovannimarchiori:
Hi @tmadlener,
running the code recently after these changes I encounter this error

calibrateECalBa...  DEBUG Input Node Name/Shape (0):
calibrateECalBa...  DEBUG       X : -1x12
calibrateECalBa...  DEBUG Output Node Name/Shape (0):
calibrateECalBa...  DEBUG       <FF><FF><FF><FF><FF><FF><FF><FF>^A : -1x1

..

calibrateECalBa...  ERROR ERROR running model inference: Invalid output name: <FF><FF><FF><FF><FF><FF><FF><FF>^A
calibrateECalBa...VERBOSE ServiceLocatorHelper::service: found service ExceptionSvc
calibrateECalBa...  ERROR Maximum number of errors ( 'ErrorMax':1) reached.
EventLoopMgr      WARNING Execution of algorithm calibrateECalBarrelClusters failed

This code used to work in the past. 
Any idea what might have gone wrong with these changes? They are quite technical so I am not sure where to start looking for a potential bug (unless the problem is that the onnx files being read are not understood if the runtime version used here is different from that used to produce the files themselves..)

Thanks for your help!

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
```diff
@@ -336,10 +338,12 @@ StatusCode PhotonIDTool::readMVAFiles(const std::string& mvaInputsFileName,
   debug() << "Output Node Name/Shape (" << m_ortSession->GetOutputCount() << "):" << endmsg;
   for (std::size_t i = 0; i < m_ortSession->GetOutputCount(); i++)
   {
-    // for old ONNX runtime version
-    // m_output_names.emplace_back(m_ortSession->GetOutputName(i, allocator));
-    // for new runtime version
+#if ORT_API_VERSION < 13
+    m_output_names.emplace_back(AllocatedStringPtr(m_ortSession->GetOutputName(i, allocator), allocDeleter).release());
+#else
     m_output_names.emplace_back(m_ortSession->GetOutputNameAllocated(i, allocator).get());
```
giovannimarchiori, line 344: should this .get() be a .release() instead?

RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -266,10 +268,11 @@ StatusCode CalibrateCaloClusters::readCalibrationFile(const std::string &calibra
   debug() << "Output Node Name/Shape (" << m_output_names.size() << "):" << endmsg;
   for (std::size_t i = 0; i < m_ortSession->GetOutputCount(); i++)
   {
-    // for old ONNX runtime version
-    // m_output_names.emplace_back(m_ortSession->GetOutputName(i, allocator));
-    // for new runtime version
+#if ORT_API_VERSION < 13
+    m_output_names.emplace_back(AllocatedStringPtr(m_ortSession->GetOutputName(i, allocator), allocDeleter).release());
+#else
     m_output_names.emplace_back(m_ortSession->GetOutputNameAllocated(i, allocator).get());
```
giovannimarchiori, line 274: should this .get() be a .release() instead?

RecFCCeeCalorimeter/src/components/PhotonIDTool.cpp
tmadlener: Yes, almost certainly. Otherwise we will have a dangling pointer to a temporary. Sorry about that. We will be leaking these pointers at the end though then, so we will probably have to cleanup in the destructor.

giovannimarchiori:
Hi @tmadlener ,
I prepared a PR with the fix in https://github.com/HEP-FCC/k4RecCalorimeter/pull/108, could you please have a look? I also added pointer cleanup in the finalize method.
If it looks good to you I will merge it, thanks!

PR #107, opened 2024-08-19
Add tools for the implmentation of noise for ALLEGRO v3

zwu0922:
Code developed under the following nightly version:
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2024-07-31

Update:
1. Set the default segmentation in "CreateCaloCells" to "ModuleThetaMerged".
2. Add new barrel geometry tool matching the segmentation of ALLEGRO v3.
3. Add a steering file and a temporary test command (to be moved to ddsim+k4run, since the test is ALLEGRO specific)

Caution: The noise filter is not ready yet.


RecCalorimeter/src/components/TubeLayerModuleThetaMergedCaloTool.h
```diff
@@ -0,0 +1,59 @@
...
+  /// Name of active volumes
+  Gaudi::Property<std::string> m_activeVolumeName{this, "activeVolumeName", "LAr_sensitive"};
+  /// Name of active layers for sampling calorimeter
+  Gaudi::Property<std::string> m_activeFieldName{this, "activeFieldName", "active_layer"};
+  /// Name of the fields describing the segmented volume
+  Gaudi::Property<std::vector<std::string>> m_fieldNames{this, "fieldNames"};
+  /// Values of the fields describing the segmented volume
+  Gaudi::Property<std::vector<int>> m_fieldValues{this, "fieldValues"};
+  /// Temporary: for use with MergeLayer tool
+  Gaudi::Property<unsigned int> m_activeVolumesNumber{this, "activeVolumesNumber", 0};
```
giovannimarchiori, line 56: @BrieucF : can we get rid of this?

BrieucF: I think so

BrieucF: Is it bothering you? It looks like we could also use this if one wants to limit the number of layer in which we apply noise. I indeed do not see a clear use case for this but why to restrict existing functionalities?

giovannimarchiori: I’m against bloatware - simpler code is easier to read, understand, develop, maintain and debug. I don’t see a use case for this, and this is a new tool so we don’t need to add the full kitchen sink back just in case. If we want to have nonoise in some layers we can have empty histograms for those.

BrieucF: That is also a valid point, agreed

zwu0922: Sorry, perhaps I am really inexperienced on this subject. I checked a recently updated tool for reference. But a Gaudi property with the same purpose is also defined there:
https://github.com/zwu0922/k4RecCalorimeter/blob/main/RecFCCeeCalorimeter/src/components/NoiseCaloCellsVsThetaFromFileTool.h#L102

Maybe we can use this function to achieve our goal?
https://github.com/key4hep/k4geo/blob/main/detectorCommon/include/detectorCommon/DetUtils_k4geo.h#L195
If so, could you please point me to any examples on how to use this function?

Thanks in advance

zwu0922: The tool will not work without providing number of ECAL layers as input. For the moment, I prefer to leave this Gaudi property here, just as what we do in all other similar tools.

giovannimarchiori: Hi @zwu0922 ,
OK, indeed I tried running the tool without specifying the number of layers and the code that should get it directly from the geometry is not working. Let's keep it like that, I will remove the non-working code in a subsequent PR

giovannimarchiori:
I think we can merge the PR as is and I will fix a few little issues in a subsequent PR:
- hardcoded systemID value
- nonworking code to be removed
- move the tool to RecFCCeeCalorimeter where other "vsTheta" tools live
- revert the defaults of CreateCaloCells (these are all configurables that we can set in python to whatever it's appropriate for FCC-ee, but let's not modify the defaults)
- move the test scripts to ddsim+k4run

PR #108, opened 2024-08-29
fix crash due to wrong treatment of pointers in ONNX-related code, and remove one photon-ID variable from decorations when not needed

giovannimarchiori:
Pointers were not treated correctly in some cases and led to a crash, see discussion in https://github.com/……HEP-FCC/k4RecCalorimeter/pull/104#discussion_r1735675543

Also, one shape parameter for photon ID was being added to clusters even when the flag for adding photon-id related variables was false.

This PR fixes both things

### DIFF
--- a/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
+++ b/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
@@ -129,9 +129,9 @@ StatusCode AugmentClustersFCCee::initialize()
       showerShapeDecorations.push_back(Form("energy_fraction_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("theta_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("phi_%s_layer_%d", detector, layer));
-      showerShapeDecorations.push_back(Form("maxcell_E_%s_layer_%d", detector, layer));
       // pi0/photon shape var only calculated in EMB
       if (m_do_photon_shapeVar && m_systemIDs[k] == systemID_EMB) {
+        showerShapeDecorations.push_back(Form("maxcell_E_%s_layer_%d", detector, layer));
         showerShapeDecorations.push_back(Form("width_theta_%s_layer_%d", detector, layer));
         showerShapeDecorations.push_back(Form("width_module_%s_layer_%d", detector, layer));
         showerShapeDecorations.push_back(Form("Ratio_E_max_2ndmax_%s_layer_%d", detector, layer));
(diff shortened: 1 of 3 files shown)


RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -169,9 +169,24 @@ StatusCode CalibrateCaloClusters::finalize()
...
+
   if (m_ortEnv)
     delete m_ortEnv;
 
+  for (std::size_t i = 0; i < m_input_names.size(); i++) {
+    if (m_input_names.at(i)) {
+      delete m_input_names.at(i);
+    }
+  }
+  m_input_names.resize(0);
```
tmadlener, line 181:
```suggestion
  for (auto& name : m_input_names) {
    delete name;
  }
```
This can be shortened to this, because:
- we loop over all elements, so a range-based for loop more clearly expresses that extent.
- `delete` will happily accept a `nullptr` (and do the right thing with it)
- the `resize` doesn't have to happen, since the whole algorithm and with it the `vector` will be destroyed very soon after this. At that point it will contain a bunch of pointers that have already been cleaned up, so it will also do the right thing automatically.

RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp
```diff
@@ -129,9 +129,9 @@ StatusCode AugmentClustersFCCee::initialize()
       showerShapeDecorations.push_back(Form("energy_fraction_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("theta_%s_layer_%d", detector, layer));
       showerShapeDecorations.push_back(Form("phi_%s_layer_%d", detector, layer));
-      showerShapeDecorations.push_back(Form("maxcell_E_%s_layer_%d", detector, layer));
       // pi0/photon shape var only calculated in EMB
       if (m_do_photon_shapeVar && m_systemIDs[k] == systemID_EMB) {
+        showerShapeDecorations.push_back(Form("maxcell_E_%s_layer_%d", detector, layer));
```
tmadlener, line 134: This seems to be an unrelated change. Should it be in this PR?

tmadlener (commented):
Thanks a lot for the fix. I have one suggestion for the cleanup code, which could be made a bit shorter and easier to read. And there seems to be one unrelated(?) change here as well?

tmadlener (approved):

tmadlener:
@kjvbrt or @BrieucF can you press the necessary buttons to merge this, if you are happy as well.

PR #113, opened 2024-09-13
fix noise filtering bug in calo digitisation, and restore defaults

giovannimarchiori:
In the calo digitisation classes, when noise filtering is on, the cells which actually get populated with noise becomes smaller and smaller event after event due to the fact that the filtered cells are removed from the cell map (from e.g. this: https://github.com/HEP-FCC/k4RecCalorimeter/blob/af83315a749bf118e9b499b75cbb3904f6d07ef6/RecCalorimeter/src/components/NoiseCaloCellsFromFileTool.cpp#L110) and never re-added at the beginning of the next event. 

This PR fixes the problem. 

Also, a PR recently merged (#107) changed some default values of the properties of CreateCaloCells. Even though those are configurables and can be modified at python level, for backward compatibility I'd rather restore the original values as mentioned here https://github.com/HEP-FCC/k4RecCalorimeter/pull/107#issuecomment-2348741613

@kjvbrt I think @BrieucF is still on vacation so could you please have a look? Since it's a bug fix this is quite urgent. When merged I can then open a follow-up PR to fix the things mentioned in https://github.com/HEP-FCC/k4RecCalorimeter/pull/107#issuecomment-2348741613

Thanks, Giovanni

### DIFF
--- a/RecCalorimeter/src/components/CreateCaloCells.cpp
+++ b/RecCalorimeter/src/components/CreateCaloCells.cpp
@@ -70,6 +70,12 @@ StatusCode CreateCaloCells::initialize() {
       error() << "Unable to create empty cells!" << endmsg;
       return StatusCode::FAILURE;
     }
+    verbose() << "Initialised empty cell map with size " << m_cellsMap.size() << endmsg;
+    // noise filtering erases cells from the cell map after each event, so we need
+    // to backup the empty cell map for later reuse
+    if (m_addCellNoise && m_filterCellNoise) {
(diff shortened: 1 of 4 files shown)


kjvbrt:
Why are you event keepend cells map across events? I'm not sure it saves any cpu/memory.

giovannimarchiori:
that code predates me but I think it is to avoid invoking at each event the segmentation code that returns max/min thetaID and number of theta and phi bins for each layer and creates an empty map with all possible cellIDs. Not sure how much is the gain in CPU

PR #115, opened 2024-10-02
Switch from Associations to Links

tmadlener:
Switch happened in https://github.com/key4hep/EDM4hep/pull/341 and without this. Things will start failing once https://github.com/key4hep/EDM4hep/pull/373 is merged.

### DIFF
--- a/RecCalorimeter/src/components/CreateTruthJet.cpp
+++ b/RecCalorimeter/src/components/CreateTruthJet.cpp
@@ -7,7 +7,7 @@
 
 // EDM4hep
 #include "edm4hep/MCParticleCollection.h"
-#include "edm4hep/MCRecoParticleAssociationCollection.h"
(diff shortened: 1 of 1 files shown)


BrieucF:
Thanks Thomas!

Shall we do the same as here: https://github.com/key4hep/key4hep-dev-utils/pull/9/files ?

giovannimarchiori:
Indeed in my local version of k4RecCalorimeter I added
+# ignore false positive no-dangling-reference warnings from GNU gcc
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-dangling-reference")
+endif()

to CMakeLists.txt to get rid of those annoying false positive from gcc 14..

tmadlener:
I made a second PR to fix that to not mix things in the git history.

PR #117, opened 2024-10-15
CMake: add key4hep config, most importantly RPATH settings for instal…

andresailer:
…led libraries fixes https://github.com/key4hep/key4hep-spack/issues/655


BrieucF:
Thanks Andre!

PR #119, opened 2024-10-23
add option to save only topoclusters with E>E=min

giovannimarchiori:
when running with noise ON the number of topoclusters is huge. This can be reduced by increasing the thresholds on S/N for the seed cell but still there are a certain number of low-E clusters (typically below 20 MeV) that one might want to get rid of to keep the output size reasonable.
This code adds a property to the topoclustering code that allows the user to set a minimum energy threshold (by default is zero)

Usage example: https://gitlab.cern.ch/gmarchio/FCC-scripts/-/blob/main/run_ALLEGRO_reco.py?ref_type=heads#L893

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
+++ b/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
@@ -111,23 +111,24 @@ StatusCode CaloTopoClusterFCCee::execute(const EventContext&) const {
     debug() << "No active cells, skipping event..." << endmsg;
     return StatusCode::SUCCESS;
   }
-  debug() << "Number of active cells: " << allCells.size() << endmsg;
+  debug() << "Number of active cells:                              : " << allCells.size() << endmsg;
 
   // On first event, create cache
   if (m_min_noise.empty()) {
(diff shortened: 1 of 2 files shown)


BrieucF:
I have no objection! @kjvbrt ok for you?

RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
```diff
@@ -157,91 +159,104 @@ StatusCode CaloTopoClusterFCCee::execute(const EventContext&) const {
...
-      counter++;
+      energy += allCells[cID];
     }
-    cluster.addToShapeParameters(deltaR / energy);
-    verbose() << "Cluster energy:     " << cluster.getEnergy() << endmsg;
-    checkTotEnergy += cluster.getEnergy();
+    verbose() << "Cluster energy:     " << energy << endmsg;
+    checkTotEnergy += energy;
+  
+    if (energy>=m_minClusterEnergy) {
```
kjvbrt, line 170: Is it possible to have inverted logic:
```
if (energy < m_minClusterEnergy) {
  continue;
}
```
Just to avoid nesting...

PR #123, opened 2024-10-29
Include "edm4hep/Constants.h" where it's needed

jmcarcell:
After https://github.com/key4hep/k4FWCore/pull/253, `DataHandle.h` will not include `edm4hep/Constants.h` (it should never have because it wasn't being used at least for the last versions) so now this include will need to happen wherever some label from `edm4hep::labels` is used.


jmcarcell:
Merging to build the nightlies

PR #125, opened 2024-11-01
Fix compiler warnings

jmcarcell:
We should enable the same warnings as in Key4hep probably, but that's left for another time. Maybe it's good if some people have a look at the warnings in case the expectations were different from what the code was doing. In particular here: https://github.com/HEP-FCC/k4RecCalorimeter/pull/125/files#diff-0341e38011283eda958d6139213699c31fe2f8e76825d208ea0691a0d3529216L274-L275, where a variable was being set to itself.


giovannimarchiori:
This looks good to me. @BrieucF, @kjvbrt : OK if I merge?

BrieucF:
Please proceed ;-)

PR #129, opened 2024-11-11
fix for ecal-hcal barrel linking in neighbour finder

giovannimarchiori:
Fixes a couple of issues pointed out by Archil Durglishvili when matching ecal-hcal
BEGINRELEASENOTES
- bug in bin calculation (shifted by 0.5) 
- avoid looping both on ecal and hcal
ENDRELEASENOTES
Checked the impact of both changes: 1st one had tiny effect on cluster energy; 2nd has no impact as expected


BrieucF:
Can you fix https://github.com/HEP-FCC/k4RecCalorimeter/actions/runs/11786172575/job/32828926005?pr=129#step:3:326 ?

giovannimarchiori:
Fixed

PR #130, opened 2024-11-13
Update path to the reference SIM file for cross-talk tests

BrieucF:
BEGINRELEASENOTES
- Update the path to the reference SIM file for cross-talk ctests

ENDRELEASENOTES


BrieucF:
See https://github.com/HEP-FCC/FCC-config/pull/220

PR #135, opened 2025-01-22
Update for changes in allegro ecal segmentation.

scott-snyder:
The allegro ecal segmentation class has been changed to properly return a position in the volume-local coordinate system.  Remove the workaround here that was dealing with what the segmentation used to return.

BEGINRELEASENOTES
- Update cell position calculation for Allegro ECal to go along with changes in k4geo.
ENDRELEASENOTES

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CellPositionsECalBarrelModuleThetaSegTool.cpp
+++ b/RecFCCeeCalorimeter/src/components/CellPositionsECalBarrelModuleThetaSegTool.cpp
@@ -42,19 +42,7 @@ StatusCode CellPositionsECalBarrelModuleThetaSegTool::initialize() {
   }
     
 
-  // Take readout bitfield decoder from GeoSvc
-  m_decoder = m_geoSvc->getDetector()->readout(m_readoutName).idSpec().decoder();
   m_volman = m_geoSvc->getDetector()->volumeManager();
-
-  // check if decoder contains "layer"
-  std::vector<std::string> fields;
-  for (uint itField = 0; itField < m_decoder->size(); itField++) {
-    fields.push_back((*m_decoder)[itField].name());
-  }
-  auto iter = std::find(fields.begin(), fields.end(), "layer");
-  if (iter == fields.end()) {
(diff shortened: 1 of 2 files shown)


scott-snyder:
Depends on PR#1384 from DD4hep (https://github.com/AIDASoft/DD4hep/pull/1384) and PR#423 from k4geo (https://github.com/key4hep/k4geo/pull/423).

scott-snyder:
Now that the k4geo change was merged, this should also get merged.  Otherwise, cell positions will be wrong.

scott-snyder:
The failures are because we're not seeing the updated version of k4geo, on which this depends.
I'll try to put in a temporary hack again to work around.

giovannimarchiori:
Hi @scott-snyder 
I don't think we need a work around. Tomorrow the nightly will have the k4geo with your changes merged in and we can rerun the tests and they should succeed - then we merge

giovannimarchiori (commented):
The changes look good to me, they are basically straightforward given the way the underlying code in k4geo (and DD4hep) has been reshuffled. 
Maybe one comment is that the release notes are a bit misleading, the reader could think that the cells were badly positioned before, which is not the case - the user will get the same positioning with this positioning tool before/after changes.

scott-snyder:
Thanks.  I tried to update the description slightly to be clearer.
I'll hold off on making any changes for now.  However, i would expect the tests against the release would still fail, since the k4geo (and dd4hep) changes wouldn't have gotten there.  (That was the last holdup in getting the k4geo change merged.)

giovannimarchiori:
Mmm ok, for k4RecCalorimeter it usually suffices that the tests are passed for the nightlies for us to accept the pr..

giovannimarchiori:
Tests for nightlies passed successfully, going to merge

giovannimarchiori:
just to confirm that with 1000 photon events (10 GeV, theta between 50 and 130 degrees) the cell positions (with the position tool) are identical before/after merging this PR, and that the hits from the readout (summing the G4 steps with same cellID) have now the proper z position

PR #145, opened 2025-03-26
SiPM emulation Gaudi algorithm

SanghyunKo:
BEGINRELEASENOTES
- New SiPM emulation algorithm using either scintillation energy deposit or the number of optical photons

ENDRELEASENOTES

Implemented SiPM emulation Gaudi algorithms based on the SimSiPM library. Each algorithm can take either the number of optical photon (from the Cherenkov channel of the dual-readout calorimeter) or scintillator energy deposit (from generic scintillator-based sampling calorimeter).

In the former case, the timing & wavelength distributions of optical photons are assumed to be available from the SD action (we cannot store time & wavelength of each optical photon due to the storage issue). The photon wavelength is determined on-the-fly using the wavelength distribution as PDF.

In the latter case, we determine the number of photons based on the energy deposit and (empirical) scintillation yield (and the optical filter if exists). The photon wavelength is generated on-the-fly in the same manner.


SanghyunKo:
Looks like CI is failing for some other reasons
```
RecFCCeeCalorimeter/CMakeFiles/k4RecFCCeeCalorimeterPlugins.dir/src/components/CreateFCCeeCaloNeighbours.cpp.o -c /k4reccalorimeter/RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
/k4reccalorimeter/RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp:14:10: fatal error: detectorSegmentations/FCCSWHCalPhiRow_k4geo.h: No such file or directory
   14 | #include "detectorSegmentations/FCCSWHCalPhiRow_k4geo.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

RecCalorimeter/tests/options/runSiPMDualReadout.py
```diff
@@ -0,0 +1,115 @@
...
+        "DRcaloSiPMreadout_scint",
+        "DRcaloSiPMreadoutRawHit",
+        "DRcaloSiPMreadoutTimeStruct",
+        "DRcaloSiPMreadoutWaveLen"
+    ],
+    OutputLevel = DEBUG
+)
+
+from Configurables import SimulateSiPMwithEdep
+sipmEdep = SimulateSiPMwithEdep("SimulateSiPMwithEdep",
```
BrieucF, line 19: Event though they are the default, can you explicitly set `inputHitCollection` here?

RecCalorimeter/src/components/SimulateSiPMwithOpticalPhoton.h
```diff
@@ -0,0 +1,104 @@
...
+private:
+  // integral function for the wavelength random generation
+  std::vector<double> integral(const std::vector<double>& wavelen, const std::vector<double>& yval) const;
+
+  // Random Number Service
+  SmartIF<IRndmGenSvc> m_randSvc;
+  Rndm::Numbers m_rndmUniform;
+
+  // Input collections
+  mutable DataHandle<edm4hep::RawCalorimeterHitCollection> m_rawHits{"DRcaloSiPMreadoutRawHit", Gaudi::DataHandle::Reader, this};
```
BrieucF, line 62: For these three lines, can you make the name of the input collection configurable from the steering file? Either as you did with `SimulateSiPMwithEdep` or with something like https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecFCCeeCalorimeter/src/components/AugmentClustersFCCee.cpp#L26 .

RecCalorimeter/src/components/SimulateSiPMwithEdep.h
```diff
@@ -0,0 +1,92 @@
...
+#include "edm4hep/CalorimeterHitCollection.h"
+#include "edm4hep/EventHeaderCollection.h"
+
+#include "k4FWCore/DataHandle.h"
+
+#include "Gaudi/Algorithm.h"
+#include "GaudiKernel/IRndmGenSvc.h"
+#include "GaudiKernel/RndmGenerators.h"
+#include "GaudiKernel/ToolHandle.h"
+
```
BrieucF, line 14: Can you add a Doxygen description of the class?

RecCalorimeter/tests/options/runSiPMDualReadout.py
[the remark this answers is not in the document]
SanghyunKo: I implemented a CTest, it seems running fine :)

SanghyunKo:
I found that we might need a decent amount of work for handling positions of the calo hits (in which we don't care at all in the current PR), but I'd prefer to move on and work on it with another PR.

atolosadelgado (commented):
Hi @SanghyunKo 

Thank you for this PR!

I have a question related to the random number service, is that the preferred way to go?  

Somehow related to that question, is this algorithm deterministic, so if the same event is fed to the algorithm, its output is also the same?

I was trying to fix some issues related to proper management of PRNG [1], but I am still not sure if this is the way, so I would like to learn :)

Regarding the sampling of the wavelength of Cherenkov photons from a flat distribution, I guess it introduces a small "Chromatic error". Could you please confirm if it is negligible? 

Thank you for your time!

[1] https://github.com/key4hep/k4RecTracker/pull/52

SanghyunKo:
Hi @atolosadelgado, thanks for the review.

Regarding the RNG, unfortunately I'm not the right person to answer, since I also took it from the VTX digitizer [[link](https://github.com/key4hep/k4RecTracker/blob/master/VTXdigi/src/VTXdigitizer.cpp#L17)]. But in general, I'd prefer to leave it to Gaudi and don't mess around with it (i.e. avoid micro-managing with TRandom3 or other RNG).

For sampling photon wavelength from PDF, yes I have already tested it. The followings are the Cherenkov channel ADC count with 10 GeV e- gun in the case of 1) all filtering and PDE emulation done in the SD action with individual `G4OpcialPhoton` 2) take PDF instead and sample the wavelength from it.

1) 
before
2) 
after

atolosadelgado:
> Hi @atolosadelgado, thanks for the review.
> 
> Regarding the RNG, unfortunately I'm not the right person to answer, since I also took it from the VTX digitizer [[link](https://github.com/key4hep/k4RecTracker/blob/master/VTXdigi/src/VTXdigitizer.cpp#L17)]. But in general, I'd prefer to leave it to Gaudi and don't mess around with it (i.e. avoid micro-managing with TRandom3 or other RNG).
> 
> For sampling photon wavelength from PDF, yes I have already tested it. The followings are the Cherenkov channel ADC count with 10 GeV e- gun in the case of 1) all filtering and PDE emulation done in the SD action with individual `G4OpcialPhoton` 2) take PDF instead and sample the wavelength from it.
> 
> 
> before 2) after

thanks for your prompt reply!

I am glad to see the impact on the cherenkov photons is small, thanks!

Regarding the PRNG, @m-fila @jmcarcell could you please confirm if Gaudi service is fine?

m-fila:
AFAIK `RndmGenSvc` and `Numbers` don't provide an interface to easily set seeds per event per algorithm which is done to be able to reproduce the results
`IRndmGenSvc` has a method `setSeeds` but this is a seed for the whole service. In principle something like obtaining a seed for algorithm/event then setting it to RndmGenSvc then populating Numbers would work but only in a single-threaded mode - in multi-threaded multiple services might change the service seed in parallel and the result won't be predictable

SanghyunKo:
Seems like key4hep/k4geo#444 broke the CI @mahmoudali2

```
Info in <TGeoManager::CloseGeometry>: Building cache...
Info in <TGeoManager::CountLevels>: max level = 8, max placements = 21600
Info in <TGeoManager::CloseGeometry>: 260244571 nodes/ 36653 volume UID's in Detector Geometry
Info in <TGeoManager::CloseGeometry>: ----------------modeler ready----------------
Compact          INFO  ++ Converted field: Successfully created field CompensatingSolenoidZminus [solenoid]
Detector         INFO  +++ Patching names of anonymous shapes....
DD4hepVolumeManager INFO  **** running plugin DD4hepVolumeManager ! 
VolumeManager    INFO   - populating volume ids - be patient ...
IDDescriptor     ERROR dd4hep: MuonSystemCollection: This ID descriptor has no field with the name: side
DD4hep_VolumeManager ERROR Exception: IDDescriptor: dd4hep: MuonSystemCollection: This ID descriptor has no field with the name: side
     dd4hep: while programming VolumeManager. Are your volIDs correct?
Traceback (most recent call last):
  File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2025-04-09/x86_64-ubuntu22.04-gcc11.4.0-opt/dd4hep/3466ee762ed148ff4f798292c90e229e53b6c0bd_develop-54lxis/bin/ddsim", line 25, in <module>
    sys.exit(RUNNER.run())
             ^^^^^^^^^^^^
  File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2025-04-09/x86_64-ubuntu22.04-gcc11.4.0-opt/dd4hep/3466ee762ed148ff4f798292c90e229e53b6c0bd_develop-54lxis/lib/python3.11/site-packages/DDSim/DD4hepSimulation.py", line 321, in run
    kernel.loadGeometry(str("file:" + os.path.abspath(compactFile)))
cppyy.gbl.std.runtime_error: void dd4hep::sim::Geant4Kernel::loadGeometry(const string& compact_file) =>
    runtime_error: DD4hep_VolumeManager: Exception: IDDescriptor: dd4hep: MuonSystemCollection: This ID descriptor has no field with the name: side
     dd4hep: while programming VolumeManager. Are your volIDs correct?
dd4hep: with plugin:DD4hepVolumeManager
dd4hep: Error interpreting XML nodes of type <plugin/>
dd4hep: Error interpreting XML nodes of type <plugins/>
dd4hep: while parsing file:/k4reccalorimeter/k4geo/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
dd4hep: with plugin:DD4hep_XMLLoader
```

SanghyunKo:
@mahmoudali2 sorry, nevermind. It looks like the issue of old nightly release...

```
Setting up the Key4hep software stack release latest-opt from CVMFS
Use the following command to reproduce the current environment: 

        source /cvmfs/sw.hsf.org/key4hep/setup.sh -r 2025-01-28
```

SanghyunKo:
Hi @lopezzot, I'm tagging you to let you know about the outcome of the digitizer development.

It seems like using the same Gaudi module (`SimulateSiPMwithEdep`) for the Cherenkov channel of IDEA o2 is doable; The hack is to set `scintYield` to 10^-6/keV = 1/GeV (so that "1" stored in `edm4hep::SimCalorimeterHit::getEnergy()` yields exactly one photo electron) and `scintDecaytime` to 0. The timing and attenuation is estimated by using the distance to the sensor from the `G4Step`, assuming that the `edm4hep::SimCalorimeterHit::getPosition()` returns the sensor position.

But be aware that you need to set `m_hitCreationMode = HitCreationFlags::DETAILED_MODE` in your SD action to store the position of each `edm4hep::CaloHitContribution` because of the [following part](https://github.com/AIDASoft/DD4hep/blob/3466ee762ed148ff4f798292c90e229e53b6c0bd/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L632-L636) in DD4hep. I suppose this is the only change needed in your code.

```c++
const Geant4HitData::Contribution& c = *ci;
int trackID = pm->particleID(c.trackID);
auto mcp = m_particles.at(trackID);
sCaloHitCont.setEnergy( c.deposit/CLHEP::GeV );
sCaloHitCont.setTime( c.time/CLHEP::ns );
sCaloHitCont.setParticle( mcp );

if ( hit_creation_mode == Geant4Sensitive::DETAILED_MODE )     {
  edm4hep::Vector3f p(c.x/CLHEP::mm, c.y/CLHEP::mm, c.z/CLHEP::mm);
  sCaloHitCont.setPDG( c.pdgID );
  sCaloHitCont.setStepPosition( p );
}
```

lopezzot:
Hi @SanghyunKo,

thanks!

One side comment. We have observed in IDEA_o2 that it is better to save as position information of each hit/fiber the inner tip position rather than the outer tip position (with outer tip I mean where the SiPM is located). If the outer tip position is used when calculating the theta and phi coordinates of showers in the calorimeter, some weird tails in the distributions are found. So for the moment we are saving in `edm4hep::SimCalorimeterHit` the inner tip position.

However, if we save for each contribution the time of arrival at the SiPM surface, this is all we need for the SiPM emulation. Also, if we follow this approach, we don't need to save positions for each contribution, because each hit corresponds to a fiber and the only position associated to the fiber is the inner tip (or outer tip). I think this would be the smallest output we need to digitize each signal.

We can chat about it whenever you want.

Cheers.

SanghyunKo:
@lopezzot No problem, I can just add a switch to use the `CaloHitContribution::getTime()` as it is rather than calculating the time based on the distance. I suppose we'll have a chance to discuss what is the most optimal definition of the calo hit position when it comes to the clustering, but for the digitizer I think a switch is all we need.

lopezzot:
Yes, agreed. I will let you know when we are ready to test the digitizer with IDEA_o2 data.

Issue #84, opened 2024-06-18
ClusterCell collection lacking cellID encondings, or including cells from heterogeneous subsystems with different CellIDEncodings


giovannimarchiori:
The clustered cell collection (https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecCalorimeter/src/components/CreateCaloClustersSlidingWindow.cpp#L45C8-L45C23) can potentially hold cells from heterogeneous subsystems which then have inconsistent CellIDEncodings. The issue and further details are extensively described and discussed in https://github.com/HEP-FCC/k4RecCalorimeter/pull/83

In addition, even if only one system is used, the cell collection attached to the cluster does not have the CellIDEncoding. This is indeed potentially a problem (analysis scripts later in the chain may need this information to properly set-up the decoder).

BrieucF:
Among the two options discussed (have similar cellID encodings for the different sub-systems and having one cluster cell collection per subsystem) I would be in favor of the latter. What do you think?

tmadlener:
Just to be sure that I understand tho goal correctly: There is a `ClusterCollection` acting as a *superset collection* (subset collection in podio terms) that contains Clusters from several subsystems that have potentially different CellID encodings. Is this correct? Or is this combined collection actually a proper collection of cloned Clusters?

Regardless of which is the case for the collection, it's necessary to get to the correct CellID encoding for each cluster in order to interpret the CellID correctly.

The main problem currently is that the combined / superset collection doesn't have any metadata set, resp. if we would set a CellID encoding for that collection we could only set one (assuming we use the usual convention to set that). However, how we can solve this depends quite a bit on whether the collection is a subset collection or whether it is a proper collection.

giovannimarchiori:
Hi @tmadlener 
What we have now is a ClusterCollection (which is not a subset collection) that contains clusters whose cells are from more than 1 system (e.g. ECAL + HCAL combined clusters), with different CellID encodings.

One could get the proper CellID by encoding in the metadata a dict of keys of type system : encoding, based on the assumption that system is always encoded in the first 4 bits of the cellID no matter the details of the readout, but that is neither elegant nor probably very robust and would maybe also break some design assumptions of podio.

[...]

PR #151, opened 2025-04-28
cleanup and generalise CaloTowerToolFCCee

giovannimarchiori:
BEGINRELEASENOTES
- Simplify code and generalise CaloTowerToolFCCee. Can now pass a list of cell collections via the cells parameters of the algorithm. Removed all dependencies on readout names and segmentations. Removed unused old code.
- Calculate SW cluster energy per subdetector and save it in subdetectorEnergies vector
- Add possibility to create links from clusters to existing input cell collections rather than creating a new output cell collections (similarly to what is done by LC / Pandora), based on the value of attachCells (true: create new output collection; false: set links to pre-existing input collections)
- If new output cell collection is created, then save in metadata the lists of input collections and corresponding system IDs, so that it's possible to retrieve proper CellID encoding for the cells in the output collection based on their system ID value
ENDRELEASENOTES

Also fixed a couple of typos.
The tool can now be called with

```
     towerTool = CaloTowerToolFCCee(outputClusters + "TowerTool",
                                   deltaThetaTower=4 * 0.009817477 / 4, deltaPhiTower=2 * 2 * pi / 1536.,
                                   thetaMin=0.0, thetaMax=pi,
                                   phiMin=-pi, phiMax=pi,
                                   cells=cells,
                                   nSubDetectors=3,
                                   OutputLevel=INFO)

```

instead of
```
     towerTool = CaloTowerToolFCCee(outputClusters + "TowerTool",
                                   deltaThetaTower=4 * 0.009817477 / 4, deltaPhiTower=2 * 2 * pi / 1536.,
                                   ecalBarrelReadoutName=inputReadouts.get("ecalBarrel", ""),
                                   ecalEndcapReadoutName=inputReadouts.get("ecalEndcap", ""),
                                   ecalFwdReadoutName=inputReadouts.get("ecalFwd", ""),
                                   hcalBarrelReadoutName=inputReadouts.get("hcalBarrel", ""),
                                   hcalExtBarrelReadoutName=inputReadouts.get("hcalExtBarrel", ""),
                                   hcalEndcapReadoutName=inputReadouts.get("hcalEndcap", ""),
                                   hcalFwdReadoutName=inputReadouts.get("hcalFwd", ""),
                                   OutputLevel=INFO)
    towerTool.ecalBarrelCells.Path = inputCells.get("ecalBarrel", "emptyCaloCells")
    towerTool.ecalEndcapCells.Path = inputCells.get("ecalEndcap", "emptyCaloCells")
    towerTool.ecalFwdCells.Path = inputCells.get("ecalFwd", "emptyCaloCells")
    towerTool.hcalBarrelCells.Path = inputCells.get("hcalBarrel", "emptyCaloCells")
    towerTool.hcalExtBarrelCells.Path = inputCells.get("hcalExtBarrel", "emptyCaloCells")
    towerTool.hcalEndcapCells.Path = inputCells.get("hcalEndcap", "emptyCaloCells")
    towerTool.hcalFwdCells.Path = inputCells.get("hcalFwd", "emptyCaloCells")

```


giovannimarchiori:
@BrieucF I need to update the branch and also to check one little thingy but I would like to have an early feedback from your side.
There are a couple of advantages:
- got rid of all dependencies on segmentation and readout
- cluster energies per subdetector calculated and saved as foreseen by EDM
- code is generic one can pass as many input cell collections as desired and if one subdetector is not included there is no need to pass the emptyCellCollection instead

BrieucF:
That is really nice Giovanni, wonderful! I will take a look asap!

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
```diff
@@ -102,110 +97,57 @@ class CaloTowerToolFCCee : public AlgTool, virtual public ITowerToolThetaModule
...
+  /// Size of the tower in phi (default: pi/314 ~ 0.01 rad)
+  Gaudi::Property<float> m_deltaPhiTower{this, "deltaPhiTower", M_PI/314, "Size of the tower in phi"};
+
+  /// Number of calorimeters (e.g. ecal/hcal/muon) for which to calculate the
+  /// total subdetector energy of the cluster and save it in the subdetectorEnergies vector
+  /// Should be equal to max(caloID) where caloID is defined as calculated in
+  /// CreatePositionedCaloCells
+  /// Barrel and endcap are considered as part of the same subdetector
+  /// If the property is 0, subdetector energy information will not be computed
+  Gaudi::Property<unsigned int> m_nSubDetectors{this, "nSubDetectors", 0, "Number of calorimeter systems"};
```
BrieucF, line 143: Do we need to provide this? Can it differ from the size of `m_cellCollections`?

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -102,198 +49,59 @@ StatusCode CaloTowerToolFCCee::finalize() {
...
-  debug() << "Input Hcal forward cell collection size: " << hcalFwdCells->size() << endmsg;
-  // Loop over a collection of calorimeter cells and build calo towers
-  if (hcalFwdCells->size()>0) {
-    CellsIntoTowers(aTowers, hcalFwdCells, fillTowersCells);
-    totalNumberOfCells += hcalFwdCells->size();
+  // Loop over input cell collections to build towers
+  for (size_t ih = 0; ih < m_cellCollectionHandles.size(); ih++) {
+    verbose() << "Processing collection " << ih << endmsg;
+    const edm4hep::CalorimeterHitCollection* coll = m_cellCollectionHandles[ih]->get();
+    debug() << "Input ell collection size: " << coll->size() << endmsg;
```
BrieucF:
```suggestion
    debug() << "Input cell collection size: " << coll->size() << endmsg;
```

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -306,116 +114,66 @@ float CaloTowerToolFCCee::theta(int aIdTheta) const {
...
+    //debug() << "Cell: x = " << cellX << " y = " << cellY << " z = " << cellZ << endmsg;
+    //debug() << "Cell: theta = " << cellTheta << " phi = " << cellPhi << endmsg;
+    //debug() << "Cell: iTheta = " << iTheta << " iPhi = " << iPhi << " iPhi(cyclic) = " << phiIndexTower(iPhi) << endmsg;
+    aTowers[iTheta][phiIndexTower(iPhi)] += cell.getEnergy() * sin(cellTheta);
+    if (fillTowersCells) {
+      clusteredCells++;
+      m_cellsInTowers[std::make_pair(iTheta, phiIndexTower(iPhi))].push_back(cell.clone());
+      int ncells = m_cellsInTowers[std::make_pair(iTheta, phiIndexTower(iPhi))].size();
+
+      if (ncells > 5)
```
BrieucF, line 171: Why that?

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -102,198 +49,59 @@ StatusCode CaloTowerToolFCCee::finalize() {
...
+  debug() << "Total number of clustered input cells: " << totalNumberOfClusteredCells << endmsg;
+
+  return totalNumberOfClusteredCells;
 }
 
 // Get the tower IDs in theta
 // aTheta Position of the calorimeter cell in theta
+// Note that the function returns an unsigned int so it
+// assumes that aTheta is <= thetaMax.
+// This is checked in CellsIntoTowers
```
BrieucF, line 92: So we are by-passing the possible finer granularity of strip layers here? I think it is the right thing to do, but just want to make sure I understand

BrieucF (commented):
Hi Giovanni, thanks a lot! I did not spot anything from a first look. Is there something in particular you have a doubt about?

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
```diff
@@ -39,61 +20,27 @@ StatusCode CaloTowerToolFCCee::initialize() {
...
+    }
   }
-  // Maximum theta & phi of the calorimeter system
-  m_phiMax = *std::max_element(listPhiMax.begin(), listPhiMax.end());
-  m_thetaMax = *std::max_element(listThetaMax.begin(), listThetaMax.end());
-  debug() << "Detector limits: phiMax " << m_phiMax << " thetaMax " << m_thetaMax << endmsg;
 
-  // very small number (epsilon) substructed from the edges to ensure correct division
-  float epsilon = 0.0001;
+  double epsilon = 0.0000;
```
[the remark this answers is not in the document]
giovannimarchiori: Hi @BrieucF 
I put this to zero to test if we really need this epsilon thing or not. I see the correct number of bins being estimated, so I would remove completely this epsilon correction, but if you think that it's better to keep it then I will put it back to 0.0001 and subtract it from thMax  - thin

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.h
giovannimarchiori: ATM yes. I am filling the subdetectorEnergies vector as is done in the LC/Pandora world: entry 0 is ECAL, 1 is HCAL, 2 is MUON, .... based on the "calorimeter type" info that I encode in the cell type.
So for instance if we are setting up a clustering algo that include ECAL barrel+endcap collections, HCAL barrel+endcap ... the size of this subdetectorEnergies does not match the number of input collections since they are split

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
giovannimarchiori: I just added a comment here but did not change the code behaviour. To answer your question: yes, cells in the strips within the same tower (say, if the tower corresponds to a big cell, 4 strip cells) are assigned the same tower ID (is that your question?)

giovannimarchiori:
> Hi Giovanni, thanks a lot! I did not spot anything from a first look. Is there something in particular you have a doubt about?

Hi Brieuc, thanks for checking. No, I just wanted to make sure that this MR would make sense before doing further development. I just replied to your comments, a couple of questions are not fully clear to me. Currently I am almost done adding a few extra features to this MR:
* option to add links from clusters to either the new output cell collection or to the input output collections (and not create a new output collection)
* in case a new output collection is created, add to the metadata of the event the mapping between systemIDs and inputCollections (or encoding) of the clustering algorithm so that one can properly decode the cellID information of these hits (see longstanding discussion never resolved in https://github.com/HEP-FCC/k4RecCalorimeter/issues/84)

giovannimarchiori:
@BrieucF I am a bit stuck when saving in the metadata the mapping between cell collection names and system IDs for the cells used for clustering:
- I wanted to pass to the tower tool also the list of system IDs (optional), and then retrieve both cell collection names and vector of systemIDs in the clustering algorithm from its tower tool. This however does not work because the towertool interface does not expose a method to return these lists, I would need to modify the interfaces
- I can add the list of systemIDs to the towertool and stop there: the configuration metadata of the job will then contain, in the settings of the towertool, information of this kind: 
`, ToolSvc.EMBCaloClustersTowerTool.calorimeterIDs = "[ 4 ]";
, ToolSvc.EMBCaloClustersTowerTool.cells = "[ 'ECalBarrelModuleThetaMergedPositioned' ]";
`
This beast however is a bit complicated to navigate
- I can add the lists of collection names and system IDs to the clustering algorithm, so that then I can create the metadata:
```
int parameters

Key                           Value 
--------------------------------------------------------------------------------
CaloClusters__inputSystemIDs  [4, 5, 8, 9]
EMBCaloClusters__inputSystemIDs[4]
EMECCaloClusters__inputSystemIDs[5]

...

std::string parameters
Key                           Value 
--------------------------------------------------------------------------------
CaloClusters__inputCellCollections[ECalBarrelModuleThetaMergedPositioned, ECalEndcapTurbinePositioned, HCalBarrelReadoutPositioned, HCalEndcapReadoutPositioned]
..
ECalBarrelModuleThetaMergedPositioned__CellIDEncoding[system:0:4,cryo:4:1,type:5:3,subtype:8:3,layer:11:8,module:19:11,theta:30:10]
...
```
But in this case, both the towertool and the clustering algorithm with have properties for what is essentially the same thing (the list of names of input cell collections) and the user might inadvertently set them differently (is there a way to set the properties of the tool from the calling algorithm even if they are private?)

So I am currently doing this last thing, setting up the algorithm and tool in python like this:
```
    towerTool = CaloTowerToolFCCee(outputClusters + "TowerTool",
                                   deltaThetaTower=4 * 0.009817477 / 4, deltaPhiTower=2 * 2 * pi / 1536.,
                                   thetaMin=0.0, thetaMax=pi,
                                   phiMin=-pi, phiMax=pi,
                                   #cells=cells,
                                   #calorimeterIDs=caloIDs,
                                   nSubDetectors=3,
                                   OutputLevel=DEBUG)

    clusterAlg = CreateCaloClustersSlidingWindowFCCee("Create" + outputClusters,
                                                      towerTool=towerTool,
                                                      nThetaWindow=windT, nPhiWindow=windP,
                                                      nThetaPosition=posT, nPhiPosition=posP,
                                                      nThetaDuplicates=dupT, nPhiDuplicates=dupP,
                                                      nThetaFinal=finT, nPhiFinal=finP,
                                                      energyThreshold=threshold,
                                                      energySharingCorrection=False,
                                                      attachCells=True,
                                                      cells=cells,
                                                      calorimeterIDs=caloIDs,
                                                      OutputLevel=DEBUG
                                                      )
    clusterAlg.clusters.Path = outputClusters
    clusterAlg.clusterCells.Path = outputClusters.replace("Clusters", "Cluster") + "Cells"
    clusterAlg.towerTool.cells = cells
    clusterAlg.towerTool.calorimeterIDs = caloIDs
    TopAlg += [clusterAlg]
```

I wonder if you have a better suggestion?
I pushed the latest version of my code so that you can see how it works at the moment. 
I look forward to your input on how best to proceed!

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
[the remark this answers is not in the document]
BrieucF: If you witness that you can get rid of it you can, maybe just leave a comment that it was removed in case this cause a problem later, it would help future debugging

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
BrieucF: Yes I see that it was there already, I am just wondering why to have this condition before to print the verbose message. Not a big deal

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
BrieucF: Yes, thanks

BrieucF:
I anticipate that the usage of the heterogeneous cells attach to the cluster in sub-sequent analysis code will likely be easier to be done using linkCollections, which you plan to create IIUC. So all good. But indeed, since there is a OneToMany from cluster to cells, it would be natural to also populate it. For this point, I think it is worth adding the needed methods to the interface, to avoid defining twice the list of system IDs in the steering file. If you do not want to modify the interface, normally you should be able to just add new methods to the `CaloTowerToolFCCee` class which inherits from the interface class, no?

giovannimarchiori:
>  If you do not want to modify the interface, normally you should be able to just add new methods to the CaloTowerToolFCCee class which inherits from the interface class, no?
I had tried adding the methods to the CaloTowerToolFCCee  but the compiler was complaining that ToolHandle<ITowerToolThetaModule> does not provide these methods

BrieucF:
Was it complaining from `CaloTowerToolFCCee` or from the sliding window clustering class, where the method is used? If the latter, I think you just have to explicitly cast it to the inherited type, without it it think it is of type `ITowerToolThetaModule`.

RecFCCeeCalorimeter/src/components/CaloTowerToolFCCee.cpp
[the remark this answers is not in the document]
giovannimarchiori: OK to be on the safe side I put back the epsilon stuff

giovannimarchiori:
yes exactly it was from the SW clustering because I was not doing the cast... 
ok if this is acceptable I will modify the code this way..

giovannimarchiori:
BTW @BrieucF I think it would make sense to rename "attachCells" to something like "createNewClusterCellCollection" - any comment?

BrieucF:
> BTW @BrieucF I think it would make sense to rename "attachCells" to something like "createNewClusterCellCollection" - any comment? I would also add the same property to the topocluster alg, so that the _hits links can be filled either with links to the original input collections, or to the new one, and in this case save the mapping systemID<->cells in metadata as I do for SW

No problem for me, but shall we tackle the topo clustering in another PR maybe? We should also coordinate with @SanghyunKo  which is now playing with topo clustering (see e.g. https://github.com/HEP-FCC/k4RecCalorimeter/issues/152) to avoid conflicts.

> In alternative, we can just remove the attachCells property, and decide what to do depending on whether the property setting the name of the new output cell container is filled or empty

I think I prefer to be explicit (keeping the attachCells property).

giovannimarchiori:
> No problem for me, but shall we tackle the topo clustering in another PR maybe? We should also coordinate with @SanghyunKo which is now playing with topo clustering (see e.g. https://github.com/HEP-FCC/k4RecCalorimeter/issues/152) to avoid conflicts.

Yes, also because the changes needed to the topo-clustering code if we don't create a new output collection are also a bit non trivial, we need to change a bit the way the various input collections are handled

> I think I prefer to be explicit (keeping the attachCells property).
OK, that's also my preferred choice.

giovannimarchiori:
Hi @BrieucF 
I think I implemented all comments, and tests are passing.
The PR is ready, I will have to fix the downstream reco script in FCC-config accordingly but it's something I can do very quickly

BrieucF:
Could you maybe compare the distribution of cluster energy and position before/after the change?

giovannimarchiori:
> Could you maybe compare the distribution of cluster energy and position before/after the change?

I checked cluster E/x/y/z and get very good agreement before/after changes

PR #154, opened 2025-05-05
Allow topoclustering algorithm to use overrided Segmentation::neighbour function

SanghyunKo:
BEGINRELEASENOTES
- Allow topoclustering algorithm to use overrided Segmentation::neighbour function

ENDRELEASENOTES

This PR allows to use of overrided `Segmentation::neighbour` function to avoid using `TTree` based neighbor map (and instead retrieves the neighboring `CellID` on-the-fly). A switch `useNeighborMap` is added to support backward compatibility.

### DIFF
--- a/RecCalorimeter/src/components/ConstNoiseTool.h
+++ b/RecCalorimeter/src/components/ConstNoiseTool.h
@@ -74,7 +74,7 @@ private:
   SmartIF<IGeoSvc> m_geoSvc;
 
   /// Decoder
-  dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder(m_systemEncoding.get());
+  dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder(m_systemEncoding);
 };
 
 #endif /* RECCALORIMETER_CONSTNOISETOOL_H */
--- a/RecCalorimeter/src/components/ConstNoiseTool.h
+++ b/RecCalorimeter/src/components/ConstNoiseTool.h
@@ -67,11 +67,14 @@ private:
   Gaudi::Property<std::vector<double>> m_detectorsNoiseOffset{
       this, "detectorsNoiseOffset", {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, "Cell noise offset in GeV"};
 
+  /// System encoding string
+  Gaudi::Property<std::string> m_systemEncoding{this, "systemEncoding", "system:4", "System encoding string"};
+
   /// Pointer to the geometry service
   SmartIF<IGeoSvc> m_geoSvc;
 
   /// Decoder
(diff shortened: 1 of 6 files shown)


RecCalorimeter/src/components/ConstNoiseTool.h
```diff
@@ -71,7 +71,7 @@ class ConstNoiseTool : public AlgTool, virtual public INoiseConstTool {
   SmartIF<IGeoSvc> m_geoSvc;
 
   /// Decoder
-  dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder("system:4");
+  dd4hep::DDSegmentation::BitFieldCoder* m_decoder = new dd4hep::DDSegmentation::BitFieldCoder("system:5");
```
giovannimarchiori, line 74: since not all calorimeters (at the moment at least) are using the same number of bits to encode the system ID, can we make this string a Gaudi property of the tool?

RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.h
```diff
@@ -119,6 +120,14 @@ class CaloTopoClusterFCCee : public Gaudi::Algorithm {
   mutable ToolHandle<INoiseConstTool> m_noiseTool{"TopoCaloNoisyCells", this};
   /// Handle for neighbours tool
   mutable ToolHandle<ICaloReadNeighboursMap> m_neighboursTool{"TopoCaloNeighbours", this};
+  // flag to use a pre-calculated neighbor map
+  Gaudi::Property<bool> m_useNeighborMap{this, "useNeighborMap", true, "use pre-calculated neighbor map"};
+  // use GeoSvc when the neighbor map is not present
+  SmartIF<IGeoSvc> m_geoSvc;
+  // name of the readout (or segmentation)
```
giovannimarchiori, line 127: can you add in the comment and in the description o the property that this is needed only when the neighbor map is not used?

RecCalorimeter/src/components/ConstNoiseTool.h
SanghyunKo: Hi, thanks for the comment. I didn't realize that the system encoding string could be different... Made it to `Gaudi::Property` as suggested.

giovannimarchiori: Thanks!

giovannimarchiori (approved):
Looks good from my side

RecFCCeeCalorimeter/tests/options/IDEA_o1_v03_reco.py
BrieucF, line 1: Hi Sanghyun, did you maybe meant to include this in a CTest?

SanghyunKo: Hi Brieuc; Yes, I'd do that, but we need to merge #145 first... (we don't have digi merged yet)

giovannimarchiori: @SanghyunKo #145 is merged... can we proceed with this PR?

PR #156, opened 2025-05-14
Add REQUIRED for find_package in CMakeLists.txt

jmcarcell:
Some of these packages are needed so they should be REQUIRED to fail immediately if they are not found.


giovannimarchiori (approved):
Approved

PR #157, opened 2025-05-15
Configurable ALLEGRO ECAL cross-talk coefficients

zwu0922:
BEGINRELEASENOTES

- To be merged only after the PR https://github.com/key4hep/k4geo/pull/458 is complete.
- Update `CreateFCCeeCaloXTalkNeighbours` to allow configurable cross-talk coefficients for the creation of calorimeter cross-talk map.
- Update the test script `runCaloXTalkNeighbours.py` accordingly.

ENDRELEASENOTES


RecFCCeeCalorimeter/tests/options/runCaloXTalkNeighbours.py
```diff
@@ -25,6 +25,10 @@
                                        systemValues=[4],
                                        activeFieldNames=["layer"],
                                        activeVolumesNumbers=[11],
+                                       xtalkCoefRadial=0.7e-2,
+                                       xtalkCoefTheta=*0.2e-2,
```
giovannimarchiori, line 29: typo: *0.2e-2 => 0.2e-2

giovannimarchiori:
Hi @zwu0922 , while you are at it, could you also consider the following changes to this code?
- replace the CELLDEBUGINFO macro with a Gaudi property, so that one does not need to recompile code to enable or disable that
-

RecFCCeeCalorimeter/tests/options/runCaloXTalkNeighbours.py
zwu0922: Typo fixed

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloXTalkNeighbours.h
```diff
@@ -65,6 +65,15 @@ class CreateFCCeeCaloXTalkNeighbours : public extends1<Service, ICaloCreateMap>
   // Theta ranges of layers in the segmented volumes
   Gaudi::Property<std::vector<std::vector<double>>> m_activeVolumesTheta{this, "activeVolumesTheta"};
 
+  // radial crosstalk coefficient, see https://indico.cern.ch/event/1368231/contributions/5904291/
+  Gaudi::Property<double> m_xtalk_coef_radial{this, "xtalkCoefRadial", 0.7e-2};
+  // thetacrosstalk coefficient
```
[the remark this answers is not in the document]
zwu0922: Typo fixed in the comment

zwu0922:
> Hi @zwu0922 , while you are at it, could you also consider the following changes to this code?
> 
>     * replace the CELLDEBUGINFO macro with a Gaudi property, so that one does not need to recompile code to enable or disable that

The DEBUGCELLINFO macro is replaced by a Gaudi property.

>     * from what I can see, the code has properties that are meant to allow this tool to be used to generate xtalk maps for all calorimeters, i.e. including also the HCAL and xtalk between ECAL and HCAL, however this is not handled at all in the C++. Basically many configurable properties are ignored. I would suggest (1) to edit the .h to add a comment that the properties starting from ecalBarrelSysId are there for future implementation of xtalk between ECAL and HCAL but currently not used; to set connectBarrels to false by default and (2)  in the C++ print out some WARNING at initialisation if connectBarrels is true to inform the user that the feature is not yet implemented.

(1) The suggested comment about properties after ecalBarrelSysId is added in the header file. m_connectBarrels is set to false by default.
(2) A warning is added in C++, in case connectBarrels is changed to true.

>     * at some point the tool could also be extended to include x-talk for the segmentation classes of the HCAL, but that should be in a separate PR

PR #158, opened 2025-05-22
Don't use ROOT typedef

tmadlener:
BEGINRELEASENOTES
- Avoid the usage of `Double_t` typedef in method signature to be less dependent on included headers

ENDRELEASENOTES

The `Double_t` typedef has pointed to `double` in ROOT for >20 years at this point and it would require to forward declare it in the header to avoid brittle builds that depend on it being included by one of the included headers. Switching to `double` seems to be the easiest solution here for me.


giovannimarchiori:
Duplicate of #160

PR #162, opened 2025-06-09
TopoCluster cell link update

jroloff:
BEGINRELEASENOTES
* Add possibility to create links from topoclusters to existing input cell collections rather than creating a new output cell collections (similarly to what is done by LC / Pandora), based on the value of m_createClusterCellCollection (true: create new output collection; false: set links to pre-existing input collections)

ENDRELEASENOTES

This PR is based on the changes from https://github.com/HEP-FCC/k4RecCalorimeter/pull/151, but extends this to apply to TopoClusters, and not just the sliding window clusters.

The tool can be called like

`    clusterAlg = CaloTopoClusterFCCee("Create" + outputClusters,
                                      cells=cells,
                                      clusters=outputClusters,
                                      clusterCells=outputClusters.replace("Clusters", "Cluster") + "Cells",
                                      neigboursTool=neighboursTool,
                                      noiseTool=noiseTool,
                                      seedSigma=seedSigma,
                                      neighbourSigma=neighbourSigma,
                                      lastNeighbourSigma=lastNeighbourSigma,
                                      minClusterEnergy=clusteringThreshold,
                                      calorimeterIDs=caloIDs,
                                      createClusterCellCollection=doCreateCells,
                                      OutputLevel=INFO)`

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
+++ b/RecFCCeeCalorimeter/src/components/CaloTopoClusterFCCee.cpp
@@ -82,6 +82,8 @@ StatusCode CaloTopoClusterFCCee::initialize() {
     return StatusCode::FAILURE;
   }
 
+
+
   // setup system decoder
   m_decoder = new dd4hep::DDSegmentation::BitFieldCoder(m_systemEncoding);
   m_indexSystem = m_decoder->index("system");
@@ -97,7 +99,10 @@ StatusCode CaloTopoClusterFCCee::execute(const EventContext&) const {
 
   // Create output collections
   edm4hep::ClusterCollection* outClusters = m_clusterCollection.createAndPut();
-  edm4hep::CalorimeterHitCollection* outClusterCells = m_clusterCellsCollection.createAndPut();
+  edm4hep::CalorimeterHitCollection* outClusterCells = nullptr;
+  if (m_createClusterCellCollection) {
+    outClusterCells = m_clusterCellsCollection.createAndPut();
+  }
 
   // Get input collection with calorimeter cells
   edm4hep::CalorimeterHitCollection* inCells = new edm4hep::CalorimeterHitCollection();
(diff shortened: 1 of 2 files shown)


giovannimarchiori:
Hi @jroloff 
overall looks good to me, but
1) there is one test failing - is it because your branch is not updated?
2) can you add, as in #151 , the writing to metadata of the cellidenconding information for the various input cell collections when m_createClusterCellCollection is true? 
Thanks!

giovannimarchiori:
Hi @jroloff 
in the test I see this error
```
EventLoopMgr        FATAL in sysInitialize(): exception with tag=Property::fromString is caught
EventLoopMgr        ERROR Property::fromString 	 Cannot convert '['DetID_SiWr_Barrel']' for property 'calorimeterIDs' in class 'CaloTopoClusterFCCee': std::invalid_argument, cannot parse '['DetID_SiWr_Barrel']' to std::vector<int,std::allocator<int> >	 StatusCode=FAILURE
```

Isn't it because you are doing
`caloIDs = ["DetID_SiWr_Barrel"]`
instead of
`caloIDs = [IDs["DetID_SiWr_Barrel"]]`
?
Moreover  - why these detector IDs? The silicon wrapper is not a calorimeter... I think you should rather use DRcalo

Note that IDEA and ALLEGRO have different subdetectors...

giovannimarchiori:
BTW - you should also rebase your branch

RecFCCeeCalorimeter/tests/options/IDEA_o1_v03_reco.py
```diff
@@ -18,6 +18,34 @@
...
+root = tree.getroot()
+IDs = {}
+for constant in root.find('define').findall('constant'):
+    if (constant.get('name') == 'DetID_VXD_Barrel' or
+        constant.get('name') == 'DetID_VXD_Disks' or
+        constant.get('name') == 'DetID_DCH' or
+        constant.get('name') == 'DetID_SiWr_Barrel' or
+        constant.get('name') == 'DetID_SiWr_Disks' or
+        constant.get('name') == 'DetID_ECAL_Barrel' or
+        constant.get('name') == 'DetID_ECAL_Endcap' or
```
giovannimarchiori, line 34: These 2 do not exist in IDEA: https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/IDEA_o1_v03/DectDimensions_IDEA_o1_v03.xml

giovannimarchiori:
Hi @jroloff 
now the code is almost good to go. Please clean up a bit more the IDEA script to remove the unnecessary (and sometimes wrong) IDs. Also, before merging, I would like you to confirm that you get the same topoclusters with the old code and after these changes. Thanks a lot!

jroloff:
@giovannimarchiori -- thanks for the comments, and I will do a quick check of the output clusters later today or tomorrow.

jroloff:
@giovannimarchiori -- just to confirm, I have compared the results with and without pandora, and before and after these changes, and the topoclusters look identical. I'm not sure why the one check is failing, but it looks like it is failing for the main branch as well, so let me know if there's anything else I need to do.

giovannimarchiori:
This PR tackles (at least as an initial reasonable workaround) - together with #151 - the problem highlighted in https://github.com/HEP-FCC/k4RecCalorimeter/issues/84 => cellID encodings of the various different collections used as input for clustering are saved in metadata and can be retrieved by downstream algorithms to decode the cell information of the ClusterCell collection (if created)

PR #163, opened 2025-06-19
Add algorithm to create links between calo hits and mc particles

giovannimarchiori:
BEGINRELEASENOTES
- Introduce a new Gaudi algorithm that can create a link collection between multiple input CalorimeterHit collections and MCParticles
ENDRELEASENOTES

Note: similar to what is done in Marlin in https://github.com/iLCSoft/MarlinReco/blob/master/Analysis/RecoMCTruthLink/src/RecoMCTruthLinker.cc
CLD reco uses a wrapper of that tool to save various links to truth in their output
For mlpf for ALLEGRO we need the calohit<->mcparticle links

### DIFF
--- a/RecCalorimeter/src/components/CreateHitTruthLinks.cpp
+++ b/RecCalorimeter/src/components/CreateHitTruthLinks.cpp
@@ -7,10 +7,7 @@ DECLARE_COMPONENT(CreateHitTruthLinks)
 
 CreateHitTruthLinks::CreateHitTruthLinks(const std::string& name, ISvcLocator* svcLoc)
     : Gaudi::Algorithm(name, svcLoc) {
-  //declareProperty("hits", m_hits, "Hits collection (input)");
-  //declareProperty("cells", m_cells, "Cell collection (input)");
   declareProperty("mcparticles", m_mcparticles, "MC particles collection (input)");
-  //declareProperty("cell_hit_links", m_cell_hit_links, "The links between cells and hits (input)");
   declareProperty("cell_mcparticle_links", m_cell_mcparticle_links, "The links between cells and MC particles (output)");
 }
 
--- a/RecCalorimeter/src/components/CreateHitTruthLinks.cpp
+++ b/RecCalorimeter/src/components/CreateHitTruthLinks.cpp
@@ -30,6 +30,7 @@ StatusCode CreateHitTruthLinks::initialize() {
     }
   }
 
+  /*
   for (const auto& col : m_cellCollections) {
     debug() << "Creating handle for input cell (CalorimeterHit) collection : " << col << endmsg;
     try {
@@ -40,6 +41,7 @@ StatusCode CreateHitTruthLinks::initialize() {
       return StatusCode::FAILURE;
     }
   }
+  */
 
   for (const auto& col : m_cell_hit_linkCollections) {
     debug() << "Creating handle for input cell-hit link collection : " << col << endmsg;
@@ -376,7 +378,7 @@ StatusCode CreateHitTruthLinks::execute(const EventContext&) const {
                         debug() << "starts_in_tracker, has_pi0, has_bs, gmother_in_calo : " <<  starts_in_tracker << " " << has_pi0 << " " <<  has_bs << " " << gmother_in_calo << endmsg;
                         if ( starts_in_tracker == 1 ) {
                           // this_Kid is a clear-cut hit-originator
-                          // this_Kid started before the calo, and is the one\
+                          // this_Kid started before the calo, and is the one
                           // the hit should be attributed to
                           remap_as_you_go[index_mcp] = index_this_Kid;
                           index_mcp = index_this_Kid;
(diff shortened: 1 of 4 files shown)


BrieucF:
> I calculate the weight of a link as the fraction of energy of a calo hit being due to a given particle, and most of my weights are close to 1. On the other hand, the weights from the marlin alg are quite close to zero...

What process is it? For particle guns, I expect weights == 1 in all cases if you go back to the primary MC particle

giovannimarchiori:
It's a multiple particle gun, in each events there are 10-15 particles in a given cone

RecCalorimeter/src/components/CreateHitTruthLinks.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+    // Loop over the G4 hits, find the associated calo hits, then loop over the G4 hit contribution to calculate contribution of each MCParticle to the calo hit
+    for (const auto &simHit : *simCalorimeterHits){
+
+      debug() << "Processing new sim hit: " << endmsg;
+      std::map< size_t , double > simHitMapEnergy ;  //  counts total energy for every MCParticle
+      int nhits = 0;
+      for (const auto &assoc : *caloHitSimCaloHitLinks)
+      {
+        if (assoc.getTo() == simHit) {
```
BrieucF, line 94: Why do you need to loop first on SimHits, then on the link? Would'nt just loop over the link work?

RecCalorimeter/src/components/CreateHitTruthLinks.cpp
```diff
@@ -0,0 +1,533 @@
...
+                    {
+                      const edm4hep::MCParticle& mother = mcparticles->at(index_mother);
+                      debug() << "shower-part mother "<<mother.id()
+                              << " gs "<<mother.getGeneratorStatus()
+                              << " dint "<<mother.isDecayedInTracker()
+                              << " bs "<<mother.isBackscatter()
+                              << " ndi "<<mother.vertexIsNotEndpointOfParent()
+                              << " npar "<<mother.getParents().size()
+                              << " pdg "<<mother.getPDG() << endmsg;
+                    }
```
BrieucF, line 209: why do you encapsulate this?

RecCalorimeter/src/components/CreateHitTruthLinks.h
```diff
@@ -0,0 +1,68 @@
...
+  virtual ~CreateHitTruthLinks();
+
+private:
+  /// List of input sim calo hit collections
+  Gaudi::Property<std::vector<std::string>> m_hitCollections{
+    this, "hits", {}, "Names of SimCalorimeterHit collections to read"};
+  /// the vector of input k4FWCore::DataHandles for the input hit collections
+  std::vector<k4FWCore::DataHandle<edm4hep::SimCalorimeterHitCollection>*> m_hitCollectionHandles;
+
+  /// List of input cell collections (not used explicitly - should it be defined anyway?)
```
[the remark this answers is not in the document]
giovannimarchiori: removed

RecCalorimeter/src/components/CreateHitTruthLinks.cpp
```diff
@@ -0,0 +1,533 @@
...
+
+StatusCode CreateHitTruthLinks::execute(const EventContext&) const {
+
+  // map from true tracks linked to hit contributions to
+  // those that really should have been linked
+  std::map< int , int > remap_as_you_go ;
+  bool doRemapping = true;  // for debug
+
+  // create cell<->particle links
+  edm4hep::CaloHitMCParticleLinkCollection* caloHitMCParticleLinkCollection = new edm4hep::CaloHitMCParticleLinkCollection();
```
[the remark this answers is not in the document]
giovannimarchiori: done

giovannimarchiori:
And these are the weights from running the test on Pythia ee->Z->qq
Screenshot 2025-06-25 at 14 46 40

BrieucF:
Thanks Giovanni, the weight we have seem reasonable, I am actually surprised by the weights obtained from CLD. Have you already propagated this to a next step to see if it does not go crazy?

giovannimarchiori:
I don't have yet a code that makes some significant use of these weights. The flattening code to create the input trees for MLPF reads these weights, I think to decide which is the "best" particle associated to a calo hit, but have no feeling of how different would the results be if these weights were changed. 
Do you think we can proceed with this PR as is?

BrieucF:
> I don't have yet a code that makes some significant use of these weights. The flattening code to create the input trees for MLPF reads these weights, I think to decide which is the "best" particle associated to a calo hit, but have no feeling of how different would the results be if these weights were changed. Do you think we can proceed with this PR as is?

If really in a rush we could, but I would prefer to sort this double loop issue (getting rid of the current unnecessary nested loop seems an easy improvement worth trying).

RecCalorimeter/src/components/CreateHitTruthLinks.cpp
giovannimarchiori: fixed

giovannimarchiori:
I removed the extra loop and checked that the results are identical. Can we now move on?

PR #167, opened 2025-07-11
Adapt fiber DRC changes from key4hep/k4geo#486

SanghyunKo:
BEGINRELEASENOTES
- Use a dedicated function to retrieve the fiber DRC rear end position when calculating the timing, which is based on the distance between the `step` and the fiber rear end (the default position now returns the front end)
- Update IDEA reconstruction config to cluster both scintillation and Cherenkov channel together instead of the individual clustering
ENDRELEASENOTES

Adapting the changes from key4hep/k4geo#486. Two main changes:
1. Since the default position of now return the front end, we need a dedicated function to retrieve the rear end position (this introduces an unavoidable geometry dependency because the timing is calculated based on the distance between the `step` and the fiber rear end. Otherwise, the timing information is meaningless if we keep using the default position).
2. Change IDEA reconstruction config file to cluster scintillation and Cherenkov channel together (as discussed in https://indico.cern.ch/event/1566115/#47-status-and-plans-for-idea-d)

### DIFF
--- a/RecCalorimeter/tests/options/runSiPMDualReadout.py
+++ b/RecCalorimeter/tests/options/runSiPMDualReadout.py
@@ -120,6 +120,7 @@ sipmOptical = SimulateSiPMwithOpticalPhoton("SimulateSiPMwithOpticalPhoton",
         0.173, 0.166, 0.158, 0.15, 0.12,
         0.05
     ],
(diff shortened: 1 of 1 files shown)


SanghyunKo:
Note: dependency on key4hep/k4geo#486 (cannot compile this PR without the k4geo update)

PR #168, opened 2025-07-21
Barrel/endcap neighbors

varnes:
BEGINRELEASENOTES

 Update CreateFCCeeCaloNeighbours to find neighbors across ECal cryostats (i.e. barrel <--> endcap)

 Also update the endcap sampling fractions in RecFCCeeCalorimeter/tests/options/ALLEGRO_o1_v03_digi_reco.py to be ones from first-principles sampling fraction calculation rather than empirical set tuned to give correct cluster energy.  This will avoid problems at the endcap/barrel ECal and HCal boundaries, though further calibration will be required to return the correct cluster energies.

ENDRELEASENOTES


RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -490,17 +497,17 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
         dd4hep::DDSegmentation::CellID volumeId = 0;
         // Get VolumeID
         (*decoder)[m_fieldNamesSegmented[iSys]].set(volumeId, m_fieldValuesSegmented[iSys]);
-        (*decoder)[m_activeFieldNamesSegmented[iSys]].set(volumeId, ilayer);
+	(*decoder)["layer"].set(volumeId, ilayer);
```
giovannimarchiori, line 500: why this change?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -490,17 +497,17 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
         (*decoder)["theta"].set(volumeId, 0);
         (*decoder)["module"].set(volumeId, 0);
         // Get number of segmentation cells within the active volume
         // numberOfCells: return Array of the number of cells in (module, theta) and the minimum theta ID.
-        auto numCells = det::utils::numberOfCells(volumeId, *moduleThetaSegmentation);
+         auto numCells = det::utils::numberOfCells(volumeId, *moduleThetaSegmentation);
```
giovannimarchiori, line 505: why the extra space?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -516,20 +523,91 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+
+	    if ((itheta == extrema[2].first) ||  (itheta == extrema[2].second)) {
+	      // find barrel cell position
+	      
+	      double eCalBarrelTheta = m_EMB_theta_lookup[ilayer];
+	      if (itheta == extrema[2].second) eCalBarrelTheta = TMath::Pi()-eCalBarrelTheta;
+	      double eCalBarrelPhi = m_EMB_phi_lookup[ilayer][imodule];
+	      
+	      for (uint iSys2 = 0; iSys2 < m_readoutNamesSegmented.size(); iSys2++) {
+		// get segmentation
```
giovannimarchiori, line 549: code does not look indented - are you using tabs instead of whitespaces?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -571,12 +649,18 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
           debug() << "Extrema[0]: " << extrema[0].first << " , " << extrema[0].second << endmsg;
           debug() << "Extrema[1]: " << extrema[1].first << " , " << extrema[1].second << endmsg;
           debug() << "Extrema[2]: " << extrema[2].first << " , " << extrema[2].second << endmsg;
+
+	  // Set up some variables that will be useful when finding neighbors
+	  // in the EM barrel.  These will be calculated the first time through
+	  // the loop
+	  float minRhoEMBarrel = -1.;
+
```
giovannimarchiori, line 657: code does not look properly indented - are you using tabs instead of whitespaces?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -587,25 +671,32 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
                 decoder->set(cellId, "rho", irho);
                 decoder->set(cellId, "z", iz);
 
+		double endcapRho = ecalEndcapTurbineSegmentation->rho(cellId);
+		if (iWheel == 2 && iz == 0 && ((endcapRho > minRhoEMBarrel) || (minRhoEMBarrel < 0.) )) {
+		  nextToEMBarrel = true;
+		}
+
+		double endcapPhi =  TMath::ATan2(ecalEndcapTurbineSegmentation->position(cellId).y(), ecalEndcapTurbineSegmentation->position(cellId).x());
+
```
giovannimarchiori, line 680: code does not look indented

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -667,15 +758,77 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+		      (*barrelDecoder)[m_fieldNamesSegmented[0]].set(barrelVolumeId, m_fieldValuesSegmented[0]);
+		      
+		      (*barrelDecoder)["system"].set(barrelVolumeId, m_ecalBarrelSysId);
+		      (*barrelDecoder)["module"].set(barrelVolumeId, iBarrelModule);
+		      (*barrelDecoder)["layer"].set(barrelVolumeId, iMatchLayer);
+		      
+		      numCells = det::utils::numberOfCells(barrelVolumeId, *moduleThetaSegmentation);
+		      
+		      if (iSide > 0) {
+			  iBarrelTheta = numCells[2];
```
giovannimarchiori, line 815: code does not look indented - replace all tabs with equivalent number of whitespaced

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -667,15 +758,77 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+		      if (minDelTheta > 0.5) continue;
+
+		      unsigned iBarrelModule = m_EMB_h_module_vs_phi[iMatchLayer]->GetBinContent(m_EMB_h_module_vs_phi[iMatchLayer]->FindBin(endcapPhi));
+		      // set volumeID correctly
+		      dd4hep::DDSegmentation::CellID barrelVolumeId = 0;
+		      (*barrelDecoder)[m_fieldNamesSegmented[0]].set(barrelVolumeId, m_fieldValuesSegmented[0]);
+		      
+		      (*barrelDecoder)["system"].set(barrelVolumeId, m_ecalBarrelSysId);
+		      (*barrelDecoder)["module"].set(barrelVolumeId, iBarrelModule);
+		      (*barrelDecoder)["layer"].set(barrelVolumeId, iMatchLayer);
```
giovannimarchiori, line 810: would be better to use m_activeFieldNamesSegmented[iSys] instead of layer and m_fieldNamesSegmented[iSys] instead of system

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -1592,4 +1745,164 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+	std::string histname = "h_EMB_modulevphi_"+std::to_string(iBarrelLayer);
+	TH1F* h_phi = new TH1F(histname.c_str(), histname.c_str(), nModulesBarrel, -TMath::Pi(), TMath::Pi());
+	for (unsigned iBarrelModule = 0; iBarrelModule < nModulesBarrel; iBarrelModule++) {
+	  // set volumeID correctly
+	  dd4hep::DDSegmentation::CellID barrelVolumeId = 0;
+	  (*barrelDecoder)[m_fieldNamesSegmented[0]].set(barrelVolumeId, m_fieldValuesSegmented[0]);
+	  
+	  (*barrelDecoder)["system"].set(barrelVolumeId, m_ecalBarrelSysId);
+	  (*barrelDecoder)["module"].set(barrelVolumeId, iBarrelModule);
+	  (*barrelDecoder)["layer"].set(barrelVolumeId, iBarrelLayer);
```
giovannimarchiori, line 1870: better to use m_activeFieldNamesSegmented[iSys] and m_fieldNamesSegmented[iSys] for layer and system

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -1592,4 +1745,164 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+	(*endcapDecoder)["side"].set(endcapCellId, iSide);
+
+	for (unsigned iECrho = 0; iECrho < numECCellsRho; iECrho++) {
+	  std::string histname_pos = "h_modulevphi_pos_"+std::to_string(iECrho);
+	  std::string histname_neg = "h_modulevphi_neg_"+std::to_string(iECrho);
+	  TH1F *h_pos=new TH1F(histname_pos.c_str(), histname_pos.c_str(), numECModules, -TMath::Pi(), TMath::Pi());
+	  TH1F *h_neg=new TH1F(histname_neg.c_str(), histname_neg.c_str(), numECModules, -TMath::Pi(), TMath::Pi());
+	  m_EMEC_h_module_vs_phi_pos.push_back(h_neg);
+	  for (unsigned iECmodule = 0; iECmodule < numECModules; iECmodule++) {
+	    //	    std::vector<Float_t> neg_phis, pos_phis, thetas;
```
giovannimarchiori, line 1806: remove commented code

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -1592,4 +1745,164 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+	  double inLocal[] = {0, 0, 0};
+	  transformMatrix.LocalToMaster(inLocal, outGlobal);
+	  double eCalBarrelPhiOffset = std::atan2(outGlobal[1], outGlobal[0]);
+	  double eCalBarrelPhi = eCalBarrelPhiOffset;
+	  double eCalBarrelTheta = moduleThetaSegmentation->theta(barrelVolumeId);
+
+	  h_phi->SetBinContent(h_phi->FindBin(eCalBarrelPhi), iBarrelModule);
+	  phis.push_back(eCalBarrelPhi);
+	  // m_EMB_theta_lookup.push_back(eCalBarrelTheta);	  
+	  //  m_EMB_phi_lookup[iBarrelLayer].push_back(eCalBarrelPhi);
```
giovannimarchiori, line 1893: remove commented code

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
varnes: In general I'm having trouble with using things like m_activeFieldNamesSegmented[iSys] since there are 2 active names for the endcap ("rho" and "z"), so that breaks the 1-1 correspondence between the names and the iSys index.  Hard-coding was a way around this, but I'm definitely open to better suggestions.

I think you're right that there's no need to hardcode "system" rather than using m_fieldNamesSegmented[iSys], so I'll fix that.

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
varnes: Fixed

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -667,15 +758,77 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+		      float minDelTheta = 1.0;
+		      		     
+		      for (unsigned iBarrelLayer = 0; iBarrelLayer < nLayersBarrel; iBarrelLayer++) {
+			double eCalBarrelTheta = m_EMB_theta_lookup[iBarrelLayer];
+			if (iSide == -1) eCalBarrelTheta = TMath::Pi()-eCalBarrelTheta;
+			float delTheta = TMath::Abs(endcapTheta-eCalBarrelTheta);
+			if (delTheta < minDelTheta) {
+			  minDelTheta = delTheta;
+			  iMatchLayer=iBarrelLayer;
+			}
```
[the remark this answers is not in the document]
varnes: Fixed

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.h
```diff
@@ -95,6 +99,20 @@ class CreateFCCeeCaloNeighbours : public extends1<Service, ICaloCreateMap> {
...
+  // Lookup tables/histograms for cell positions in the barrel and endcap EM calorimeters
+  std::vector< TH1F* > m_EMB_h_module_vs_phi;
+  std::vector< std::vector<Float_t>> m_EMB_phi_lookup;
+  std::vector<Float_t> m_EMB_theta_lookup;
+
+  std::vector< TH1F* > m_EMEC_h_module_vs_phi_pos;
+  std::vector< TH1F* > m_EMEC_h_module_vs_phi_neg;
+  std::vector< std::vector<Float_t>> m_EMEC_pos_phi_lookup;
+  std::vector< std::vector<Float_t>> m_EMEC_neg_phi_lookup;
+  std::vector<Float_t> m_EMEC_theta_lookup;
```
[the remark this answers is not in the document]
varnes: There is already a small comment on this at line 103.  Are you looking for more detail?  I can add something like:

These are pre-filled in initialize_lookups() to avoid having to find positions of cells within the double-loop over barrel and endcap cells when finding neighbors across the barrel/endcap boundary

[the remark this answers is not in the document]
giovannimarchiori: I would like a succinct description of the format of these vectors: what does their index correspond to? what do they contain, in which format?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
giovannimarchiori: you can still leave the code as it was before with m_activeFieldNamesSegmented[iSys], add a comment that for the endcap segmentation due to the two active names this option is ignore, and hardcode z and rho for the endcap encoder, without touching the barrel code

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.h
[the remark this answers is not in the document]
giovannimarchiori: thanks!

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
giovannimarchiori: it shouldn't be the case. in the python, the order of the m_activeFieldNamesSegmented entries should correspond to the order of the systems in m_fieldValuesSegmented{this, "systemValues", {4, 8, 9}};
if the system value for the endcap is after (before) the barrel there, then this should also be the case in m_activeFieldNamesSegmented.
Can you point me to the python script are you referring to?

Also, I would suggest that you update the default values of m_fieldValuesSegmented and so on so that the script by default produces the neighbour maps for the whole detector, including ecal endcap?
I mean

```
Gaudi::Property<std::vector<std::string>> m_fieldNamesSegmented{this, "systemNames", {"system", "system", "system"}};
Gaudi::Property<std::vector<int>> m_fieldValuesSegmented{this, "systemValues", {4, 8, 9}};
Gaudi::Property<std::vector<std::string>> m_activeFieldNamesSegmented{
      this, "activeFieldNames", {"layer", "layer", "layer"}};
Gaudi::Property<std::vector<unsigned int>> m_activeVolumesNumbersSegmented{
      this, "activeVolumesNumbers", {11, 13, 37}};
```

varnes: I'm guessing that you're right -- the fact that I wrote the code with both the "rho" and "z" field names rather than just "layer" may have been due to a misunderstanding on my part.  I'll see if I can rewrite it to be more similar to the barrel (as you say, we can resume the conversation in a few days so I have time to sort this out).

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -666,12 +763,82 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
                       decoder->set(newCellId, "module", newModule + ibmod / 1);
-                      decoder->set(newCellId, "layer", newLayer);
-                      neighborsList.push_back(newCellId);
+                      decoder->set(newCellId, m_activeFieldNamesSegmented[iSys], newLayer);
+                      neighboursList.push_back(newCellId);
+                    }
+                  }
+                }
+                // add neighboring cells in the EM barrel calorimeter, if relevant
+                if (nextToEMBarrel) {
```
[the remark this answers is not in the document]
varnes: Yes, now fixed.

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -524,10 +531,86 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
-                id, det::utils::neighbours_ModuleThetaMerged(*moduleThetaSegmentation, *decoder,
-                                                             {m_activeFieldNamesSegmented[iSys], "module", "theta"},
-                                                             extrema, id, m_includeDiagonalCells)));
+            auto neighboursList = det::utils::neighbours_ModuleThetaMerged(
+                *moduleThetaSegmentation, *decoder, {m_activeFieldNamesSegmented[iSys], "module", "theta"}, extrema, id, m_includeDiagonalCells);
+
+            // check if we are on an edge in theta, and add neighbours in the
+            // endcap calorimeter if so
+
+            if ((itheta == extrema[2].first) || (itheta == extrema[2].second)) {
```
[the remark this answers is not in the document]
varnes: Yes, now fixed.

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -1592,4 +1759,165 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+        std::vector<Float_t> phis, thetas;
+        unsigned nModulesBarrel = (numCells[0] - 1) * moduleThetaSegmentation->mergedModules(iBarrelLayer);
+        std::string histname = "h_EMB_modulevphi_" + std::to_string(iBarrelLayer);
+        TH1F* h_phi = new TH1F(histname.c_str(), histname.c_str(), nModulesBarrel, -TMath::Pi(), TMath::Pi());
+        for (unsigned iBarrelModule = 0; iBarrelModule < nModulesBarrel; iBarrelModule++) {
+          // set volumeID correctly
+          dd4hep::DDSegmentation::CellID barrelVolumeId = 0;
+          (*barrelDecoder)[m_fieldNamesSegmented[0]].set(barrelVolumeId, m_fieldValuesSegmented[0]);
+
+          (*barrelDecoder)[m_fieldNamesSegmented[iSys]].set(barrelVolumeId, m_ecalBarrelSysId);
```
giovannimarchiori, line 1885: idem - should 1st line be removed?

RecFCCeeCalorimeter/src/components/CreateFCCeeCaloNeighbours.cpp
```diff
@@ -666,12 +763,82 @@ StatusCode CreateFCCeeCaloNeighbours::initialize() {
...
+                      if (minDelTheta > 0.5)
+                        continue;
+
+                      unsigned iBarrelModule = m_EMB_h_module_vs_phi[iMatchLayer]->GetBinContent(
+                          m_EMB_h_module_vs_phi[iMatchLayer]->FindBin(endcapPhi));
+                      // set volumeID correctly
+                      dd4hep::DDSegmentation::CellID barrelVolumeId = 0;
+                      (*barrelDecoder)[m_fieldNamesSegmented[0]].set(barrelVolumeId, m_fieldValuesSegmented[0]);
+
+                      (*barrelDecoder)[m_fieldNamesSegmented[iSys2]].set(barrelVolumeId, m_ecalBarrelSysId);
```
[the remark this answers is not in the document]
varnes: Right, must have been a leftover from when I was developing the code.  Now removed (same applies to line 1883)

PR #174, opened 2025-09-23
Update tests

giovannimarchiori:
BEGINRELEASENOTES
- Several old tests that were relying on running Geant within dd4sim and were using the legacy io started to fail in recent nightlies. This MR fixes these issues, updating those tests to use ddsim for event generation and simulation, and to use IOSvc for input/output
ENDRELEASENOTES


jmcarcell:
The alma9 test says `killed`, probably it is running out of memory, but I'm not sure what changed from before.

PR #175, opened 2025-09-24
Reduce the output size when running ALLEGRO_o1_v03_digi_reco

jmcarcell:
BEGINRELEASENOTES
- Reduce the output size when running ALLEGRO_o1_v03_digi_reco

ENDRELEASENOTES

The output of this test is currently 261 MB, which makes it hard to display if it fails on CI. After this change it is 6.7 MB.


giovannimarchiori (approved):
LGTM

Giovanni

PR #178, opened 2025-09-27
Making methods of ICalibrateCaloHitsTool const.

scott-snyder:
Make ICalibrateCaloHitsTool::calibrate const.
For now (until the interface class is changed) we retain the non-const overloads and have them call the const versions.

BEGINRELEASENOTES
- First step to make interfaces of ICalibrateCaloHitsTool const.
ENDRELEASENOTES


giovannimarchiori (approved):
LGTM

PR #179, opened 2025-09-27
Making ICalorimeterTool::prepareEmptyCells const

scott-snyder:
Make ICalorimeterTool::prepareEmptyCells const.
For now (until the interface class is changed) we retain the non-const overloads.

Also prepare for IGeoSvc::getDetector returning a const pointer.

BEGINRELEASENOTES
- First step to make interfaces of ICalorimeterTool const.
ENDRELEASENOTES


giovannimarchiori (approved):
LGTM

PR #180, opened 2025-09-27
Fix dangling reference warnings.

scott-snyder:
This warning is currently suppressed in the cmake files. But this is a warning which one really would not want to suppress --- better to rearrange the code to avoid it.  So go ahead and fix the instances of that which we get in this package.

BEGINRELEASENOTES
- Fix dangling reference compilation warnings.
ENDRELEASENOTES


giovannimarchiori:
Hi @scott-snyder,
does this PR fix all dangling reference warnings from k4RecCalorimeter? IIRC in the past there were false positives by gcc (https://stackoverflow.com/questions/78759847/gcc-14-possibly-dangling-reference-to-a-temporary-warning-or-not-depending-on). If we don't have warnings anymore, should we remove  -Wno-dangling-reference from cmake/Key4hepConfig.cmake in this package?

PR #202, opened 2026-02-20
Add SiPM digitizer which uses fired cells as calohitcontributions

lopezzot:
BEGINRELEASENOTES
- Added a new SiPM digitizer that takes as input calohitcontributions representing photo-electrons and their time of arrival at SiPMs. Added a Python config file to use with k4geo IDEA_o2 output files.

ENDRELEASENOTES

The IDEA_o2 full-simulation now stores the time of arrival of photo-electrons (fired cells) as calohitcontributions associated with simcalohits. Each simcalohit represents a fiber. Contributions represent bunches of photo-electrons contributing to the total signal for the same fiber.
This PR includes a digitizer that uses the SimSiPM library and takes as input the output file from the IDEA_o2 simulation.

I believe the same approach could be used for the SiPM digitization of the crystal em-calorimeter of IDEA_o2.

### DIFF
--- a/RecCalorimeter/src/components/SimulateSiPMwithContrib.cpp
+++ b/RecCalorimeter/src/components/SimulateSiPMwithContrib.cpp
@@ -5,7 +5,12 @@
 DECLARE_COMPONENT(SimulateSiPMwithContrib)
 
 SimulateSiPMwithContrib::SimulateSiPMwithContrib(const std::string& aName, ISvcLocator* aSvcLoc)
-    : Gaudi::Algorithm(aName, aSvcLoc) {}
+    : Gaudi::Algorithm(aName, aSvcLoc) {
+    
+  // Questo collega la stringa della Property al DataHandle
+  declareProperty("inputHitCollection", m_simHits, "input calo collection name");
+  declareProperty("outputHitCollection", m_digiHits, "output calo collection name");
+}
 
 StatusCode SimulateSiPMwithContrib::initialize() {
   StatusCode sc = Gaudi::Algorithm::initialize();
@@ -60,7 +65,7 @@ StatusCode SimulateSiPMwithContrib::initialize() {
 }
 
 StatusCode SimulateSiPMwithContrib::execute(const EventContext&) const {
-  const edm4hep::SimCalorimeterHitCollection* scintHits = m_scintHits.get();
+  const edm4hep::SimCalorimeterHitCollection* scintHits = m_simHits.get();
   edm4hep::CalorimeterHitCollection* digiHits = m_digiHits.createAndPut();
 
   // loop through the hits (each hit corresponds to a fiber)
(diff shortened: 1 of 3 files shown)


lopezzot:
Apologies for the delay @SanghyunKo @BrieucF. I switched to `SimplePde` and fixed a small bug in the creation of empty digi hits.

I checked some distributions from the new digitized output, and they look good. For instance, time-of-arrival distributions from SiPMs reading out the tube calorimeter of `IDEA_o2`:
image
image

I believe we can merge now.

RecCalorimeter/src/components/SimulateSiPMwithContrib.cpp
```diff
@@ -0,0 +1,144 @@
...
+    // Using only analog signal (ADC conversion is still experimental)
+    const sipm::SiPMAnalogSignal anaSignal = m_sensor->signal();
+
+    // if the signal never exceeds the threshold, it will return -1.
+    const double integral =
+        std::max(0., anaSignal.integral(m_gateStart, m_gateL, m_thres));           // (intStart, intGate, threshold)
+    const double toa = std::max(0., anaSignal.toa(m_gateStart, m_gateL, m_thres)); // (intStart, intGate, threshold)
+    
+    // Do not save digi hit if the integral is 0, i.e. no signal from electronics
+    if(integral < 0.01) continue;
```
SanghyunKo: Just one last comment about this newly added line: can we make this zero-suppression threshold configurable?

lopezzot: mmm... yes I think we can. If I remember correctly SimSiPM always assumes that the signal from one converted photon corresponds to integral value 1 (arbitrary units). If this is the case, we can easily add a threshold. Do you remember if any tunable parameter of SimSiPM would lead to a different gain? i.e. integral not equal to 1 for 1 converted photon?

SanghyunKo: I do see that there is a gain parameter in the SiPM property, but for the moment it's not exposed to users and always fixed to 1.0: https://github.com/EdoPro98/SimSiPM/blob/main/include/SiPMProperties.h#L119
(I don't know whether this is intended or not, but given the design I was wondering that this might be something other than 1.0 in the future)

lopezzot: eh... this is exactly what I feared. If the gain is tunable, then the zero-suppression threshold should be automatically tuned as well:  basically, they are the same value. If we introduce a zero-suppression threshold here and then expose the tunable gain in SimSiPM, things might go wrong.

lopezzot: However, as long as the threshold here is only a zero-suppression threshold, there is not much to tune, right? We just have to discard the digital hits with an integral equal to 0. Not much to tune... 
The problem comes if we also want the possibility to apply 1-suppression, 2-suppression, n-suppression...

SanghyunKo: I think we'll have to revisit the zero-suppression threshold anyway if we start including background overlays and conducting occupancy studies. This should be done in the near future, so I don't think we can live with the 1-photon threshold very long...

Anyway, I'd prefer to make it a parameter and delegate it to the user's responsibility rather than fixing it in the source code. Any user modifying this parameter should have some level of understanding of what he/she is dealing with.

lopezzot: The user might understand what the parameter does here, but I doubt he/she will know how it relates to `gain`in the sipm library. However, ok I will add it.

lopezzot: Done.

SanghyunKo (approved):

PR #252, opened 2026-04-20
Implement TubeLayerModuleThetaCaloTool::indexer().

scott-snyder:
Implement indexer() for the ALLEGRO ecal barrel, using IDMapIndexer. Also add a simple test.

BEGINRELEASENOTES
- Implement indexer() for the ALLEGRO ecal barrel, using IDMapIndexer.
ENDRELEASENOTES

### DIFF
--- a/RecFCCeeCalorimeter/src/components/TubeLayerModuleThetaCaloTool.cpp
+++ b/RecFCCeeCalorimeter/src/components/TubeLayerModuleThetaCaloTool.cpp
@@ -1,4 +1,5 @@
 #include "TubeLayerModuleThetaCaloTool.h"
+#include "RecCaloCommon/IDMapIndexer.h"
 
 // dd4hep
 #include "DD4hep/Detector.h"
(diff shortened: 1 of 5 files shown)


RecFCCeeCalorimeter/src/components/TubeLayerModuleThetaCaloTool.cpp
```diff
@@ -7,12 +8,16 @@
...
 
+#include <stdexcept>
+
 
 DECLARE_COMPONENT(TubeLayerModuleThetaCaloTool)
 
 
 StatusCode TubeLayerModuleThetaCaloTool::collectCells(std::vector<CellID>& cells) const
 {
+  cells.reserve (2041344);
```
giovannimarchiori, line 19: what happens if the number of cells is greater than 2041344 (if we change granularity or z extent of the ECAL barrel for instance)? will the indexer be expanded or will the code fail?

scott-snyder: Things will still work; memory allocation may just be a bit less efficient (as usual with std::vector::reserve).

PR #253, opened 2026-04-24
Upstream muon collider calo digis.

aloeliger:
This is the second of some amount smaller PRs, designed to split up PR https://github.com/key4hep/k4Reco/pull/50 into more manageable pieces. This PR includes changes to calorimeter digis used by the muon collider project in k4Reco, but being redirected here (see: https://github.com/key4hep/k4Reco/pull/51#issuecomment-4286409207). The original work was done by @samf25. I have renamed `CaloDigi` to` RecCaloDigi` to try and fit what I guess is the naming theme here, but I am entirely flexible on this point. The CMake file was added by me (although written by Claude, my CMake is still rudimentary at best) based on what seems to be the CMake architecture of this repository, with each distinct directory having it's own file. This CMake file works for me and compiles completely on lxplus using the latest nightly release.

As before, this is largely being opened for discussion to start the process of muon collider forks, if the ultimate PR requires a selected subset of these changes, that is something that can be discussed. This PR is being opened in draft because this is not strictly a key4hep repository, and I am not sure that this is a dependency the muon collider would like to add.

@tmadlener @madbaron @samf25, FYI

BEGINRELEASENOTES
- Add muon collider calo digi work originally from k4Reco
ENDRELEASENOTES

### DIFF
--- a/RecCaloDigi/CMakeLists.txt
+++ b/RecCaloDigi/CMakeLists.txt
@@ -1,0 +1,58 @@
+################################################################################
+# Package: RecCaloDigi
+################################################################################
+
+gaudi_add_library(RecCaloDigi
+  SOURCES
+    src/CalorimeterHitType.cc
+    src/RealisticCaloDigi.cc
+    src/RealisticCaloReco.cc
+  LINK
+    Gaudi::GaudiKernel
+    EDM4HEP::edm4hep
+    k4FWCore::k4FWCore
+    k4FWCore::k4Interface
+    DD4hep::DDCore
+    DD4hep::DDRec
+    ROOT::Core
+    ROOT::Hist
+    CLHEP::Random
+)
+
+target_include_directories(RecCaloDigi PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+)
+
+# Concrete algorithm plugins — loaded at runtime by Gaudi
+gaudi_add_module(k4RecCaloDigiPlugins
+  SOURCES
+    src/FilterDoubleLayerHits.cc
+    src/RealisticCaloDigiScinPpd.cc
+    src/RealisticCaloDigiSilicon.cc
+    src/RealisticCaloRecoScinPpd.cc
+    src/RealisticCaloRecoSilicon.cc
+  LINK
+    RecCaloDigi
+    Gaudi::GaudiKernel
+    EDM4HEP::edm4hep
+    k4FWCore::k4FWCore
+    k4FWCore::k4Interface
+    DD4hep::DDCore
+    DD4hep::DDRec
+    ROOT::Core
+    ROOT::Hist
+    CLHEP::Random
+)
+
+target_include_directories(k4RecCaloDigiPlugins PRIVATE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
+)
+
+install(TARGETS RecCaloDigi k4RecCaloDigiPlugins
+  EXPORT k4RecCalorimeterTargets
+  RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT bin
+  LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}" COMPONENT shlib
+                                                COMPONENT dev
+  PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+)
--- a/RecCaloDigi/include/CalorimeterHitType.h
+++ b/RecCaloDigi/include/CalorimeterHitType.h
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2020-2024 Key4hep-Project.
+ *
+ * This file is part of Key4hep.
+ * See https://key4hep.github.io/key4hep-doc/ for further info.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef CalorimeterHitType_h
+#define CalorimeterHitType_h 1
+
+#include <iostream>
+
+/** Helper class for decoding/encoding lcio::CalorimeterHit types for the ILD
+ *  detector. The encoding is: caloType + 10 * caloID + 1000 * layout + 10000 * layerNum <br>
+ *  (see enums: CaloType, CaloID and Layout for possible values).<br>
+ *  Example usage: <br>
+ *  <pre>
+ *     lcio::CalorimeterHit* cHit = .... ;
+ *
+ *     // set the type (e.g. in digitization )
+ *     cHit->setType( CHT( CHT::em ,  CHT::ecal , CHT::plug , 12 ) ) ;
+ *
+ *     ...
+ *
+ *     CHT cht = cHit->getType() ;
+ *
+ *     //   sum energies for electromagentic, hadronic and tailcatcher:
+ *     if( cht.is( CHT::em ) )
+ *          e_em +=  cHit->getEnergy() ;
+ *     else
+ *       if ( cht.is(CHT::had ) )
+ *          e_had += cHit->getEnergy() ;
+ *       else
+ *          e_muon += cHit->getEnergy() ;
+ *
+ *     // use only EcalPlug hits:
+ *     if( cht.is( CHT::ecal) && cht.is( CHT::plug) )
+ *
+ *     // get the layer number (e.g. for calibration or clustering)
+ *     unsigned l = cht.layer() ;
+ *     // or directly :
+ *     unsigned l = CHT(  cHit->getType() ).layer()  ;
+ *
+ *     // detailed print:
+ *     std::cout <<  CHT(  cHit->getType() ) << std::endl ;
+ *
+ *  </pre>
+ *
+ *  F.Gaede, DESY, 12/2008
+ */
+
+class CHT {
+public:
+  /** calorimeter types */
+  enum CaloType { em = 0, had = 1, muon = 2 };
+
+  /** calo ids - specific to ILD */
+  enum CaloID { unknown = 0, ecal = 1, hcal = 2, yoke = 3, lcal = 4, lhcal = 5, bcal = 6 };
+
+  /** calo layout / subdetector  */
+  enum Layout { any = 0, barrel = 1, endcap = 2, plug = 3, ring = 4 };
+
+  /** C'tor for initialization from CalorimeterHit::getType()  */
+  CHT(int type) : m_type(type) {}
+
+  /** C'tor  for encoding the calo type inforamtion  */
+  CHT(CaloType c, CaloID n, Layout l, unsigned lay)
+      : m_type(c * fCaloType + n * fCaloID + l * fLayout + lay * fLayer) {}
+
+  /** calorimeter type: CHT::em , CHT::had, CHT::muon */
+  CaloType caloType() const { return (CaloType)(m_type % fCaloID); }
+
+  /** calo ID - see enum CaloID for allowed values */
+  CaloID caloID() const { return (CaloID)((m_type % fLayout) / fCaloID); }
+
+  /** calo layout - see enum layout for allowed values */
+  Layout layout() const { return (Layout)((m_type % fLayer) / fLayout); }
+
+  /** calo layer of hit  */
+  unsigned layer() const { return unsigned(m_type) / fLayer; }
+
+  bool is(CaloType t) const { return caloType() == t; }
+
+  bool is(CaloID n) const { return caloID() == n; }
+
+  bool is(Layout l) const { return layout() == l; }
+
+  /** automatic conversion to int */
+  operator int() const { return m_type; }
+
+  /** explicit conversion to int */
+  int toInt() const { return m_type; }
+
+protected:
+  int m_type;
+
+  static const int fCaloType = 1;
+  static const int fCaloID   = 10;
+  static const int fLayout   = 1000;
+  static const int fLayer    = 10000;
+};
+
+/** detailed string for calo type */
+std::ostream& operator<<(std::ostream& os, const CHT& cht);
+
+/** Return Layout based on the collection name, e.g. if name contains tolower("endcap") CHT::endcap is returned. In case no known layout
+    is found, CHT::any is returned.*/
+CHT::Layout layoutFromString(const std::string& name);
+
+/** Return caloID based on the collection name, e.g. if name contains tolower("HCal") CHT::hcal is returned. In case no known type
+    is found, CHT::unknown is returned.*/
+CHT::CaloID caloIDFromString(const std::string& name);
+
+/** Return caloType from string, e.g. if name contains tolower("Had") CHT::had is returned. In case no known type
+    is found, CHT::em is returned.*/
+CHT::CaloType caloTypeFromString(const std::string& name);
+
+#endif
--- a/RecCaloDigi/include/FilterDoubleLayerHits.h
+++ b/RecCaloDigi/include/FilterDoubleLayerHits.h
@@ -1,0 +1,95 @@
+#ifndef FilterDoubleLayerHits_h
+#define FilterDoubleLayerHits_h 1
+
+#include <k4FWCore/Transformer.h>
+#include "k4Interface/IGeoSvc.h"
+#include <GaudiKernel/ITHistSvc.h>
+
+#include "DDRec/SurfaceManager.h"
+#include <edm4hep/TrackerHitPlaneCollection.h>
+#include <edm4hep/TrackerHitPlane.h>
+
+#include <string>
+
+#include <TH1.h>
+
+
+
+/** Utility processor that removes tracker hits in double layers if they don't have
+ *  a corresponding close-by hit in the other sublayer.
+ *  Pairs of considered layers are configurable and extracted from the cellID word.
+ *  Works for all four lcio hit classes.
+ *
+ *  @parameter InputCollection name of the hit collection with (Sim)TrackerHits/(Sim)CalorimeterHits
+ *  @parameter OutputCollections ( ColName  StartLayer EndLayer )
+ *
+ * @author N. Bartosik, INFN Torino, S. Ferraro
+ * @date  17 June 2020
+ * @version $Id: $
+ */
+
+struct FilterDoubleLayerHits : public k4FWCore::Transformer<edm4hep::TrackerHitPlaneCollection(
+  const edm4hep::TrackerHitPlaneCollection &)> {
+
+protected:
+
+  static const size_t NHITS_MAX = 10000000;
+
+  struct SensorPosition{
+    unsigned int layer;
+    unsigned int side;
+    unsigned int ladder;
+    unsigned int module;
+
+    bool operator<(const SensorPosition& rhs) const {
+      return std::tie(layer, side, ladder, module) < std::tie(rhs.layer, rhs.side, rhs.ladder, rhs.module);
+    }
+  };
+
+  /// Double layer cut struct
+  struct DoubleLayerCut{
+    unsigned int layer0 ;
+    unsigned int layer1 ;
+    double dPhi_max ;
+    double dTheta_max ;
+  };
+
+
+ public:
+  FilterDoubleLayerHits(const std::string& name, ISvcLocator* svcLoc) ;
+
+  /** Called at the begin of the job before anything is read.
+   * Use to initialize the processor, e.g. book histograms.
+   */
+  virtual StatusCode initialize() ;
+
+  /** Called for every event - the working horse.
+   */
+  edm4hep::TrackerHitPlaneCollection operator()(const edm4hep::TrackerHitPlaneCollection& inputTrackerHitCollection) const override;
+
+  /** Called after data processing for clean up.
+   */
+  virtual StatusCode finalize();
+
+
+ protected:
+  dd4hep::rec::Vector2D globalToLocal(long int cellID, const dd4hep::rec::Vector3D& posGlobal, dd4hep::rec::ISurface** surf) const;
+
+  Gaudi::Property<std::string> m_subDetName{this, "SubDetectorName", "Vertex", "Name of sub detector"};
+  Gaudi::Property<bool> m_fillHistos{this, "FillHistograms", false, "Whether to fill diagnostic histograms"};
+  Gaudi::Property<double> m_dtMax{this, "DeltaTimeMax", -1.0, "Maximum time difference between hits in a doublet [ns]"};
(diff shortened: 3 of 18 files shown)


tmadlener (commented):
I don't have enough permissions here to actually launch workflows and I also didn't yet have time to actually look into the code in a lot of detail. I am very happy that this pulls over the *RealisticCalo{Digi,Reco}* family of processors though.

Having said that, we had quite some discussion in the [Key4hep meeting on March 31](https://indico.cern.ch/event/1669859/) and one of the key outcomes was that we should try to avoid algorithm names like *realistic* or *detailed* but rather go for more descriptive algorithm names. I don't have an immediate idea for this family of algorithms, unfortunately.

tmadlener:
Question (most likely to @samf25): Are these direct translations of the one from https://github.com/iLCSoft/MarlinReco/tree/master/CaloDigi/Realistic, or have there been MuonCollider specific changes applied to these?

madbaron:
I'm pretty sure these are 1:1 translations

RecCaloDigi/include/CalorimeterHitType.h
```diff
@@ -0,0 +1,129 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef CalorimeterHitType_h
+#define CalorimeterHitType_h 1
+
+#include <iostream>
+
+/** Helper class for decoding/encoding lcio::CalorimeterHit types for the ILD
```
tmadlener, line 24: @jmcarcell we already have this in k4GaudiPandora, right? Can we somehow make this available to both without having to copy things?

On a bigger picture, this looks like something that we should potentially revisit (also @BrieucF) as it is very much geared towards CALICE like detectors, resp. ILC / CLIC conventions and might fall apart with other CellID encoding strings.

Hence, also connected to https://github.com/key4hep/k4geo/issues/568

RecCaloDigi/include/FilterDoubleLayerHits.h
```diff
@@ -0,0 +1,95 @@
...
+ *  @parameter InputCollection name of the hit collection with (Sim)TrackerHits/(Sim)CalorimeterHits
+ *  @parameter OutputCollections ( ColName  StartLayer EndLayer )
+ *
+ * @author N. Bartosik, INFN Torino, S. Ferraro
+ * @date  17 June 2020
+ * @version $Id: $
+ */
+
+struct FilterDoubleLayerHits : public k4FWCore::Transformer<edm4hep::TrackerHitPlaneCollection(
+  const edm4hep::TrackerHitPlaneCollection &)> {
```
tmadlener, line 32: This looks like it is more tracking related (and then should probably go to https://github.com/key4hep/k4RecTracker).

Without having looked into any details about its usage yet, why is it in this PR?

RecCaloDigi/include/RealisticCaloDigi.h
```diff
@@ -0,0 +1,118 @@
...
+    /** Called for every run.
+     */
+    std::tuple<edm4hep::CalorimeterHitCollection, 
+               edm4hep::CaloHitSimCaloHitLinkCollection> operator()(
+              const edm4hep::SimCalorimeterHitCollection& inputSim,
+              const edm4hep::EventHeaderCollection& headers) const; 
+  
+    /** Called after data processing for clean up.
+     */
+    StatusCode finalize();
```
tmadlener, line 53:
```suggestion
```
This doesn't seem to do anything (in the implementation file).

RecCaloDigi/include/RealisticCaloDigi.h
```diff
@@ -0,0 +1,118 @@
...
+    // integration result types
+    using integr_res = std::pair<float,float>;
+    using integr_res_opt = std::optional<integr_res>;
+    using integr_function = std::function<integr_res_opt(const edm4hep::SimCalorimeterHit*)>;
+
+   virtual float EnergyDigi(float energy, float event_correl_miscalib) const;
+   virtual integr_res_opt Integrate( const edm4hep::SimCalorimeterHit * hit ) const;
+   
+   integr_res_opt StandardIntegration( const edm4hep::SimCalorimeterHit * hit ) const ;
+   integr_res_opt ROCIntegration( const edm4hep::SimCalorimeterHit * hit ) const ;
```
tmadlener, line 69: All of these functions can just take the `edm4hep::SimCalorimeterHit` by value or const reference. There is no need to turn it into a pointer (which was probably done because in LCIO you only get pointers).

Additionally (also due to where this comes from likely), these functions should probably be camel cased to follow our usual conventions.

RecCaloDigi/include/RealisticCaloDigi.h
```diff
@@ -0,0 +1,118 @@
...
+  protected:
+  
+    // energy scales we know about
+    enum { MIP, GEVDEP, NPE };
+    // integration result types
+    using integr_res = std::pair<float,float>;
+    using integr_res_opt = std::optional<integr_res>;
+    using integr_function = std::function<integr_res_opt(const edm4hep::SimCalorimeterHit*)>;
+
+   virtual float EnergyDigi(float energy, float event_correl_miscalib) const;
```
tmadlener, line 65: I don't think this is used in a way that requires it to be `virtual`. All inheriting classes implement the virtual functions below.
```suggestion
   float EnergyDigi(float energy, float event_correl_miscalib) const;
```

RecCaloDigi/include/RealisticCaloReco.h
```diff
@@ -0,0 +1,72 @@
...
+  
+    /** Called for every run.
+     */
+    std::tuple<edm4hep::CalorimeterHitCollection, 
+               edm4hep::CaloHitSimCaloHitLinkCollection> operator()(
+              const edm4hep::CaloHitSimCaloHitLinkCollection& inputLinks) const; 
+  
+    /** Called after data processing for clean up.
+     */
+    StatusCode finalize();
```
tmadlener, line 48:
```suggestion
```
Doesn't do anything special, so it can go.

RecCaloDigi/include/RealisticCaloReco.h
```diff
@@ -0,0 +1,72 @@
...
+
+
+
+ protected:
+
+  float getLayerCalib( int ilayer ) const;
+  virtual float reconstructEnergy(const edm4hep::CalorimeterHit* hit, int layer) const = 0;  // to be overloaded, technology-specific
```
tmadlener, line 55:
```suggestion
  virtual float reconstructEnergy(const edm4hep::CalorimeterHit& hit, int layer) const = 0;  // to be overloaded, technology-specific
```
Similar to the `Digi` part. No need to pass pointers here if we can do const references

RecCaloDigi/include/RealisticCaloReco.h
```diff
@@ -0,0 +1,72 @@
...
+
+  // parameters
+  // Grouping of calo layers
+  Gaudi::Property<std::vector<int>> m_calLayers{this, "calibration_layergroups", {}, "Grouping of calo layers"};
+  // Calibration coefficients for layers groups
+  Gaudi::Property<std::vector<float>> m_calibrCoeff{this, "calibration_factorsMipGev", {}, "Calibration coefficients (MIP->shower GeV) of layers groups"};
+  // Cell ID layer string
+  Gaudi::Property<std::string> m_encodingStringVariable{this, "EncodingStringParameterName", "GlobalTrackerReadoutID", "The name of the DD4hep constant that contains the Encoding string for tracking detectors"};
```
tmadlener, line 63: The description was likely already off in Marlin, but this should almost certainly be the encoding string for the calorimeter?

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+  edm4hep::CalorimeterHitCollection newcol;
+  edm4hep::CaloHitSimCaloHitLinkCollection relcol;
+
+  CHT::CaloType cht_type = caloTypeFromString(m_calo_type);
+  CHT::CaloID   cht_id   = caloIDFromString(m_calo_id);
+  CHT::Layout   cht_lay  = layoutFromString(m_calo_layout);
+
+  std::string initString;
+  initString = m_geoSvc->constantAsString(m_encodingStringVariable.value());
+  dd4hep::DDSegmentation::BitFieldCoder bitFieldCoder(initString);  // check!
```
tmadlener, line 109: This can be moved to `initialize` and then we keep the `bitFieldCoder` as a member variable. We don't have to construct this for every event.

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+struct MCC {
+  float energy {0.f};
+  float time {0.f};
+};
+
+RealisticCaloDigi::RealisticCaloDigi(const std::string& name, ISvcLocator* svcLoc) : MultiTransformer(name, svcLoc, 
+    { KeyValues("inputHitCollections", {"SimCalorimeterHits"}),
+      KeyValues("inputHeaderCollections", {"EventHeader"}) },
+    { KeyValues("outputHitCollections", {"CalorimeterHits"}),
+     KeyValues("outputRelationCollections", {"CaloHitLinks"}) }) {}
```
tmadlener, line 35: These can now be single `KeyValue` to make it impossible to mis-configure from the python side.

E.g.
```cpp
KeyValue("inputHitCollection", "SimCalorimeterHits")
```

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+
+  std::string initString;
+  initString = m_geoSvc->constantAsString(m_encodingStringVariable.value());
+  dd4hep::DDSegmentation::BitFieldCoder bitFieldCoder(initString);  // check!
+
+  debug() << "Number of elements = " << inputSim.size() << endmsg;
+  // loop over input hits
+  for (int j=0; j < inputSim.size(); ++j) {
+    edm4hep::SimCalorimeterHit simhit0 = inputSim.at( j );
+    edm4hep::SimCalorimeterHit *simhit = &simhit0;
```
tmadlener, line 115: This will become unnecessary once the interfaces are adapted to take `const edm4hep::SimCalorimeterHit&`.

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+
+    // deal with energy integration and timing aspects
+    auto integrationResult = Integrate(simhit);
+    if( ! integrationResult.has_value() ) {
+      continue;
```
tmadlener, line 120: Here a log message would be good to have to make it a bit easier to trace what is happening. Not sure if this is a `warning` because it should in principle not happen, or whether this is somewhat command and `debug` would be enough.

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+	    newhit.setEnergy( energyDig );
+      
+	    int layer = bitFieldCoder.get(simhit->getCellID(), "layer");
+	    newhit.setType( CHT( cht_type, cht_id, cht_lay, layer ) );
+
+      debug() << "orig/new hit energy: " << simhit->getEnergy() << " " << newhit.getEnergy() << endmsg;
+
+      edm4hep::MutableCaloHitSimCaloHitLink rel = relcol.create();
+      rel.setTo(simhit0);
+      rel.setFrom(newhit);
```
tmadlener, line 141:
```suggestion
      rel.set(simhit0);
      rel.set(newhit);
```
The original here likely also predates this functionality (introduced in https://github.com/AIDASoft/podio/pull/257). Also not strictly necessary to change this, but in general if links have distinct types `To` and `From` are optional.

RecCaloDigi/src/RealisticCaloReco.cc
```diff
@@ -0,0 +1,92 @@
...
+#include <algorithm>
+#include <assert.h>
+#include <cmath>
+
+using namespace std;
+
+RealisticCaloReco::RealisticCaloReco(const std::string& name, ISvcLocator* svcLoc) : MultiTransformer(name, svcLoc, 
+  { KeyValues("inputLinkCollections", {"CaloHitLinks"}) },
+  { KeyValues("outputHitCollections", {"CalorimeterHitsRec"}),
+   KeyValues("outputRelationCollections", {"CaloHitLinksRec"}) }) {}
```
tmadlener, line 20: similar to the `Digi` part. This can use `KeyValue` now.

RecCaloDigi/src/RealisticCaloReco.cc
```diff
@@ -0,0 +1,92 @@
...
+
+//-----------------------------------------------------------------------------------------------
+
+std::tuple<edm4hep::CalorimeterHitCollection, 
+           edm4hep::CaloHitSimCaloHitLinkCollection> RealisticCaloReco::operator()(
+           const edm4hep::CaloHitSimCaloHitLinkCollection& inputLinks) const {
+  // * Reading Collections of digitised calorimeter Hits *
+  std::string initString;
+  initString = m_geoSvc->constantAsString(m_encodingStringVariable.value());
+  dd4hep::DDSegmentation::BitFieldCoder bitFieldCoder(initString);  // check!
```
tmadlener, line 43: Can also go to a member variable and then be populated once in `initialize`.

RecCaloDigi/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    RecCaloDigi
+    Gaudi::GaudiKernel
+    EDM4HEP::edm4hep
+    k4FWCore::k4FWCore
+    k4FWCore::k4Interface
+    DD4hep::DDCore
+    DD4hep::DDRec
+    ROOT::Core
+    ROOT::Hist
+    CLHEP::Random
```
tmadlener, line 45: In principle many of these dependencies are transitively pulled in via `RecCaloDigi`, but I have nothing against being explicit here.

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+
+  // unit in which threshold is specified
+  if (m_threshold_unit.value().compare("MIP") == 0){
+    m_threshold_iunit=MIP;
+  } else if (m_threshold_unit.value().compare("GeV") == 0){
+    m_threshold_iunit=GEVDEP;
+  } else if (m_threshold_unit.value().compare("px") == 0){
+    m_threshold_iunit=NPE;
+  } else {
+    error() << "could not identify threshold unit. Please use \"GeV\", \"MIP\" or \"px\"! Aborting." << endmsg;
```
tmadlener, line 57: This does not abort yet. It would have to return a `StatusCode::FAILURE` for that.

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+  // deal with timing calculations  
+  std::map<std::string, integr_function> integrations = {
+    {"Standard", std::bind(&RealisticCaloDigi::StandardIntegration, this, _1)},
+    {"ROC", std::bind(&RealisticCaloDigi::ROCIntegration, this, _1)}
+  };  
+  auto findIter = integrations.find( m_integration_method ) ;
+  if(integrations.end() == findIter) {
+    error() << "Could not guess timing calculation method!" << endmsg;
+    error() << "Available are: Standard, ROC. Provided: " << m_integration_method << endmsg;
+    error() << "Aborting..." << endmsg;
```
tmadlener, line 72: also not aborting

RecCaloDigi/src/RealisticCaloDigi.cc
```diff
@@ -0,0 +1,302 @@
...
+  }
+  m_integr_function = findIter->second;
+  
+  // check if parameters are correctly set for the ROC integration
+  if("ROC" == m_integration_method) {
+    if(m_fast_shaper == 0.0f || m_slow_shaper == 0.0f) {
+      error() << "Fast/slow shaper parameter(s) not set. Required for ROC integration!" << endmsg;
+      error() << "Aborting..." << endmsg;
```
tmadlener, line 80: and once more, not aborting.

tmadlener (commented):
Thanks a lot for starting this (and apologies up-front for once again being a bit pedantic and requesting quite a few changes).

CI is failing due to a sign comparison warning at the moment. Additionally, after #237 this will likely need some `clang-format` formatting to comply again.

There are a few general themes in the comments, but in general this looks pretty good to me already. The one thing that we will need to fix is the naming. Still no really good idea, as a basis for discussion:
- `RealisticCaloDigi` (base class) -> `ElectronicsSignalDigi` (maybe include a `Base` suffix as well?)
- `RealisticCaloReco` (base class) -> `ElectronicsSignalReco`

(and then similar for the others). **This is all up for discussion, these are just some initial proposals**, maybe others have better ideas.

Finally, there is quite some documentation about this here: https://github.com/iLCSoft/MarlinReco/tree/master/CaloDigi/Realistic/doc After all my proposed changes that is probably no longer accurate (or even less accurate than it is now), so my suggestion would be to bring that over in a separate PR, because it's also not entirely clear to me where that should go.

madbaron:
> The one thing that we will need to fix is the naming. Still no really good idea, as a basis for discussion:
> 
> * `RealisticCaloDigi` (base class) -> `ElectronicsSignalDigi` (maybe include a `Base` suffix as well?)
> * `RealisticCaloReco` (base class) -> `ElectronicsSignalReco`
> 
> (and then similar for the others).

Personally I'm not 100% clear on why the second step is referred to as reco. I overall find it a bit misleading, since all this algorithm does is converting an absolute MIP scale into a particle level energy calibration.

Perhaps one could refer these two as:
* BaseDigitiser
* BaseCellEnergyCalibrator
or something along these lines?

madbaron:
I've opened https://github.com/HEP-FCC/k4RecCalorimeter/pull/281 based on aloeliger:k4_muonColliderRebase_caloDigis with updates to address @tmadlener's comments to get this finalised.

I'll reply to the comments above, but suggest to pick up any follow-ups in the new PR.
This one can be closed.

RecCaloDigi/include/CalorimeterHitType.h
madbaron: You're right that this is a copy. For now I've synced our copy to be **byte-identical** to the one in k4GaudiPandora rather than leaving two versions drifting apart.

I'd rather not attempt the actual de-duplication in this PR, because "where should it live" feels like a question  that deserves discussion in an issue and to be tackled separately.

On the bigger picture: agreed, and the file rather advertises the problem. It still describes itself in LCIO terms, and the encoding is a fixed decimal packing (`caloType + 10*caloID + 1000*layout + 10000*layerNum`), so it hard-codes room for exactly 4 calo types and 100 layouts. 
That is not going to survive contact with an arbitrary cellID encoding.

RecCaloDigi/include/FilterDoubleLayerHits.h
madbaron: Agreed, and removed from this PR: it was swept in by accident when the package was assembled.
It does belong in k4RecTracker rather than here.

RecCaloDigi/include/RealisticCaloDigi.h
madbaron: Done throughout: `Integrate`, `StandardIntegration`, `ROCIntegration`, the `integr_function` alias and `reconstructEnergy` all take `const&` now, and the `SimCalorimeterHit* simhit = &simhit0;` alias in `operator()` is gone with them. Also camelCased them as you suggested.

While converting the loops I switched the `for (int j = 0; j < coll.size(); ++j)` patterns to range-based `for`, which is what removes the `-Wsign-compare` warnings CI was failing on.

RecCaloDigi/include/RealisticCaloDigi.h
madbaron: Done. I also dropped `virtual` from `Integrate` in the same commit. It has exactly the same property (nothing overrides it, it's a one-line thunk to `m_integr_function`), so it seemed odd to fix one and not the other.

RecCaloDigi/include/RealisticCaloReco.h
madbaron: This one turned out to be more than a stale description, thanks for flagging this.

Both algorithms defaulted to `GlobalTrackerReadoutID`, and nothing in our steering overrides it, so that default is what has actually been running. In MAIA the tracker's `layer` sits at bits 7–12, which in the calo encoding is the bottom of `module`.

So the digitiser has been writing the module number into the `CHT` layer field, and the calibration step has been feeding it to `getLayerCalib()`. It went unnoticed because our configs use a single calibration group, so every layer maps to the same coefficient. The bug is invisible until someone actually uses the layer-group machinery, at which point hits get the wrong coefficient. (And if a decoded value ever exceeded the group range, `getLayerCalib` returns 0 with the `assert` compiled out under `-DNDEBUG`, i.e. silent zero-energy hits.)

Default is now `GlobalCalorimeterReadoutID` with a corrected description, matching the split k4GaudiPandora already makes. I also added a one-line check in `initialize()` that the encoding actually has a `layer` field, so a future mismatch is a configuration error rather than a per-event exception.

RecCaloDigi/src/RealisticCaloDigi.cc
madbaron: Done in both algorithms. It's a member now, built once in `initialize()`. A value member works cleanly here since `BitFieldCoder` is default-constructible and copy-assignable and `get()` is `const`, so no pointer needed.

Side effect worth noting: a bad encoding string now fails at configuration time instead of on the first event, which is probably a better failure mode.

RecCaloDigi/src/RealisticCaloDigi.cc
madbaron: Done for all seven slots across the two algorithms. I checked the generated Configurables to be sure it took effect. The properties are now scalar `DataObjID` rather than `std::vector<DataObjID>`, so passing a list from Python is a type error instead of being silently accepted.

This does rename the properties (`inputHitCollections` -> `inputHitCollection` etc.), so it's a breaking change for existing steering.

RecCaloDigi/src/RealisticCaloDigi.cc
madbaron: I tried this and backed it out. The suggested form didn't compile against podio 1.7 / EDM4hep 1.1:

```
error: static assertion failed: Argument type is ambiguous, can't determine link direction
  ... set(T) [with T = edm4hep::MutableCalorimeterHit; FromT = edm4hep::CalorimeterHit;
              ToT = edm4hep::SimCalorimeterHit]
```

I passed this on to Claude, which said:
The templated `set()` dispatches on exact type — its body only tests `is_same_v<T, FromT>` and `is_same_v<T, ToT>`, both immutable handles — so a `Mutable` handle satisfies the `requires` clause and then falls through to the `static_assert`. Since these algorithms only ever link objects they just created, *every* call site is a mutable handle, so using `set()` would mean an explicit `edm4hep::CalorimeterHit(...)` conversion everywhere. 

That felt like it cost more readability than it bought, so I've left `setFrom`/`setTo`.
Let me know if I got this wrong!

RecCaloDigi/src/RealisticCaloDigi.cc
madbaron: Added, as `debug()`.  I think that branch fires whenever every contribution falls outside the integration window, and our MAIA config runs a deliberately tight window to reject out-of-time BIB, so it's the expected path for a large fraction of hits rather than an anomaly.

I also gave the threshold rejection the same treatment (it was the other silent drop) and added a one-line per-event summary: `Digitised N of M hits (X outside the integration window, Y below threshold)`.

RecCaloDigi/CMakeLists.txt
madbaron: Kept them explicit

RecCaloDigi/src/RealisticCaloDigi.cc
tmadlener: Fixed with https://github.com/AIDASoft/podio/pull/1014

PR #257, opened 2026-05-13
Use IOSvc instead of PodioDataSvc and PodioOutput

jmcarcell:
BEGINRELEASENOTES
- Use IOSvc instead of PodioDataSvc and PodioOutput, Following https://github.com/key4hep/k4FWCore/pull/392

ENDRELEASENOTES


giovannimarchiori (approved):
Excellent, thanks @jmcarcell !

PR #258, opened 2026-05-20
Add CaloCellIndexerSvc.

scott-snyder:
For efficiency, we want to be able to identify cells via a dense index; this is done by ICaloIndexer objects.  Add a service to manage these objects: given a detector ID, or set of detector IDs, it returns an ICaloIndexer object appropriate for that combination.

BEGINRELEASENOTES
- Add CaloCellIndexerSvc
ENDRELEASENOTES


giovannimarchiori (approved):
Excellent, thanks!

PR #261, opened 2026-06-10
Support additional cluster calibration features

Dr-Stubborn:
BEGINRELEASENOTES

- Updated `CalibrateCaloClusters` to support calibration ONNX models with either the legacy `nLayers + 1` inputs or the extended `nLayers + 4` inputs.
- Added support for extra cluster-position features used by the newer calibration models:
  - cluster `theta`
  - fractional `theta / deltaTheta`
  - fractional `phi / deltaPhi`
- Kept backward compatibility with existing 12-input models by deriving the expected feature layout from the ONNX input shape.
- Ensured the raw cluster energy is always placed as the final model input, matching both legacy and extended training feature orders.
- Improved calibration model shape validation messages to report both accepted input widths and the actual model input width.
- Verified the updated algorithm with the local layer-theta workflow, including `--doTopoClustering true --calibrateClusters true`.

ENDRELEASENOTES

### DIFF
--- a/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
+++ b/RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
@@ -18,7 +18,19 @@
 
 #include <onnxruntime_cxx_api.h>
 
+#include <algorithm>
+#include <cmath>
+
 #include "OnnxruntimeUtilities.h"
+
+namespace {
+constexpr double kDeltaThetaCalo = 0.009817477 / 4.0;
+constexpr double kDeltaPhiCalo = 2.0 * 3.14159265358979323846 / 1536.0;
+
+float pythonModuloOne(double value) {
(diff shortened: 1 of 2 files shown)


RecFCCeeCalorimeter/src/components/CalibrateCaloClusters.cpp
```diff
@@ -18,8 +18,20 @@
...
 #include <onnxruntime_cxx_api.h>
 
+#include <algorithm>
+#include <cmath>
+
 #include "OnnxruntimeUtilities.h"
 
+namespace {
+constexpr double kDeltaThetaCalo = 0.009817477 / 4.0;
+constexpr double kDeltaPhiCalo = 2.0 * 3.14159265358979323846 / 1536.0;
```
giovannimarchiori: I don't like to have these numbers hardcoded, I would rather suggest to add two configurable properties to the algorithm.

PR #262, opened 2026-06-11
extend AugmentClustersFCCee to segmentations different from ECAL barrel

giovannimarchiori:
BEGINRELEASENOTES
- Generalize AugmentClustersFCCee so that the decorations with the per-layer barycenters and energy fractions can be saved also for cluster of cells from segmentations different from those used in the ALLEGRO ECAL barrel
ENDRELEASENOTES


varnes (approved):
Aside from an extra line of whitespace introduced at line 340 in the revised file, these changes look fine to me.

PR #271, opened 2026-07-03
IDEA o2 seeded ECAL clustering algorithms

SanghyunKo:
BEGINRELEASENOTES
- Add "seeded" ECAL clustering algorithms (`CaloDrivenClusterSeeding`, `TrackDrivenClusterSeeding`, `ClusterSeedMerging`, and `ClusterSeedGrower`)
- Add CI test for the IDEA o2 reconstruction chain

ENDRELEASENOTES

Adds the calorimeter reconstruction chain for the IDEA o2 dual-readout calorimeter (segmented-crystal ECAL + capillary-tube HCAL).

#### CreateOpticalCaloCells (RecCalorimeter)
Turns optical-photon SimCalorimeterHits into CalorimeterHits (energy = signal × calibration, position from the SimHit) and links each cell to its truth energy-deposit SimHit. Truth matching uses the full cellID by default; `maskCherenkovForTruthLink` drops the `cherenkov` field so the crystal ECAL's scint and Cherenkov readings link to their shared deposit, while the tube HCAL keeps its distinct scint/Cherenkov cells.

#### Seeded ECAL clustering (RecFCCeeCalorimeter)
A seed → merge → grow topological clustering for the segmented crystal ECAL:
  - CaloDrivenClusterSeeding — seeds from calorimeter energy (two thresholds + neighbour counting).
  - TrackDrivenClusterSeeding — seeds from tracks extrapolated to the ECAL.
  - ClusterSeedMerging — merges calo/track seeds within a distance.
  - ClusterSeedGrower — grows the merged seeds into clusters (shared logic in ClusterSeedingBase.h).

#### CI
Runs the chain (digitize → seed → merge → grow → truth links) on a ddsim barrel wedge and checks clusters are produced. The reco config is dual-mode: CI wedge under `IDEA_O2_CI=1`, full detector otherwise.

This PR requires key4hep/k4geo#611, key4hep/k4RecTracker#86 to pass the CI test.

### DIFF
--- a/RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
+++ b/RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
@@ -1,0 +1,90 @@
+#include "CreateOpticalCaloCells.h"
+
+// EDM4hep
+#include "edm4hep/CalorimeterHit.h"
+
+// k4FWCore
+#include <k4FWCore/MetadataUtils.h>
+
+// DD4hep
+#include "DDSegmentation/BitFieldCoder.h"
+
+// STL
+#include <stdexcept>
+#include <unordered_map>
+
+DECLARE_COMPONENT(CreateOpticalCaloCells)
+
+CreateOpticalCaloCells::CreateOpticalCaloCells(const std::string& name, ISvcLocator* svcLoc)
+    : MultiTransformer(name, svcLoc,
+                       {KeyValue("OpticalHits", "SimCalorimeterHitCollection"),
+                        KeyValue("TruthHits", "SimCalorimeterHitCollection")},
+                       {KeyValue("OutputCollection", "CalorimeterHitCollection"),
+                        KeyValue("OutputLinks", "CaloHitSimCaloHitLinkCollection")}) {}
+
+StatusCode CreateOpticalCaloCells::initialize() {
+  info() << name() << ": calibration constant = " << m_calibConst.value() << endmsg;
+
+  if (m_maskCherenkov.value()) {
+    // Locate the 'cherenkov' field in the input cellID encoding to build its mask
+    const std::string opticalKey = inputLocations("OpticalHits")[0];
+    auto encoding = k4FWCore::getCellIDEncoding(opticalKey, this);
+    if (!encoding.has_value()) {
+      error() << "maskCherenkovForTruthLink is set but the cellID encoding of '" << opticalKey
+              << "' could not be retrieved" << endmsg;
+      return StatusCode::FAILURE;
+    }
+    dd4hep::DDSegmentation::BitFieldCoder decoder(encoding.value());
+    try {
+      m_cherenkovMask = decoder[m_cherenkovField.value()].mask();
+    } catch (const std::exception& e) {
+      error() << "cellID field '" << m_cherenkovField.value() << "' not found in encoding '" << encoding.value()
+              << "': " << e.what() << endmsg;
+      return StatusCode::FAILURE;
+    }
+    info() << "Cherenkov masking enabled: optical->truth match ignores field '" << m_cherenkovField.value()
+           << "' (mask 0x" << std::hex << m_cherenkovMask << std::dec << ")" << endmsg;
+  }
+  return StatusCode::SUCCESS;
+}
+
+uint64_t CreateOpticalCaloCells::truthKey(uint64_t cellID) const {
+  return m_maskCherenkov.value() ? (cellID & ~m_cherenkovMask) : cellID;
+}
+
+std::tuple<edm4hep::CalorimeterHitCollection, edm4hep::CaloHitSimCaloHitLinkCollection>
+CreateOpticalCaloCells::operator()(const edm4hep::SimCalorimeterHitCollection& opticalHits,
+                                   const edm4hep::SimCalorimeterHitCollection& truthHits) const {
+  edm4hep::CalorimeterHitCollection outputHits;
+  edm4hep::CaloHitSimCaloHitLinkCollection outputLinks;
+
+  // Index the truth (energy-deposit) hits by their (optionally cherenkov-masked) cellID
+  std::unordered_map<uint64_t, edm4hep::SimCalorimeterHit> truthHitMap;
+  truthHitMap.reserve(truthHits.size());
+  for (const auto& truthHit : truthHits) {
+    truthHitMap[truthKey(truthHit.getCellID())] = truthHit;
+  }
+
+  // Digitize the optical hits and link each cell to its truth deposit
+  for (const auto& opticalHit : opticalHits) {
+    if (opticalHit.getEnergy() <= 0) {
+      continue; // skip cells with no signal
+    }
+    auto caloHit = outputHits.create();
+    caloHit.setCellID(opticalHit.getCellID());
+    caloHit.setPosition(opticalHit.getPosition());
+    caloHit.setEnergy(opticalHit.getEnergy() * m_calibConst.value());
+
+    auto truthIt = truthHitMap.find(truthKey(opticalHit.getCellID()));
+    if (truthIt != truthHitMap.end()) {
+      auto hitLink = outputLinks.create();
+      hitLink.setFrom(caloHit);
+      hitLink.setTo(truthIt->second);
+    }
+  }
+
+  debug() << "Digitized " << opticalHits.size() << " optical hits -> " << outputHits.size() << " cells with "
+          << outputLinks.size() << " truth links" << endmsg;
+
+  return std::make_tuple(std::move(outputHits), std::move(outputLinks));
+}
--- a/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
+++ b/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
@@ -1,0 +1,174 @@
+#include "CaloDrivenClusterSeeding.h"
+
+#include "edm4hep/CalorimeterHit.h"
+#include "edm4hep/MutableCluster.h"
+
+#include "DD4hep/Detector.h"
+#include "DDSegmentation/Segmentation.h"
+
+#include "ClusterSeedMerging.h" // for enum
+
+#include <unordered_map>
+#include <vector>
+#include <queue>
+#include <cstdint>
+
+// ============================================================
+//  CaloDrivenClusterSeeding
+// ============================================================
+
+CaloDrivenClusterSeeding::CaloDrivenClusterSeeding(const std::string& name,
+                                                   ISvcLocator*      svcLoc)
+    : ClusterSeedingBase(name, svcLoc,
+                       {KeyValues("InputCaloHitCollections", {})},
+                       {KeyValue("OutputSeedsA", "CaloDrivenSeedsA"),
+                        KeyValue("OutputSeedsB", "CaloDrivenSeedsB")}) {}
+
+StatusCode CaloDrivenClusterSeeding::initialize() {
+  return ClusterSeedingBase::initialize();
+}
+
+// ------------------------------------------------------------
+
+std::tuple<edm4hep::ClusterCollection, edm4hep::ClusterCollection>
+CaloDrivenClusterSeeding::operator()(
+    const std::vector<const edm4hep::CalorimeterHitCollection*>& caloHitColls) const {
+
+  edm4hep::ClusterCollection seedsA;
+  edm4hep::ClusterCollection seedsB;
+
+  const float thrA = m_seedEnergyThresholdA.value();
+  const float thrB = m_seedEnergyThresholdB.value();
+
+  // ------------------------------------------------------------------
+  // Step 1: Accumulate ECAL scintillation depth-0 energy per cellID.
+  //         Also keep a representative CalorimeterHit per cellID so we
+  //         can attach it to the output cluster.
+  // ------------------------------------------------------------------
+  std::unordered_map<uint64_t, float> energyMap;
+  ClusterSeedingBase::Hitmap hitMap;
+
+  // create lookup maps by looping over all input collections and hits
+  for (const auto* coll : caloHitColls) {
+    if (!coll)
+      continue;
+
+    for (const auto& hit : *coll) {
+      const uint64_t cellID = hit.getCellID();
+
+      if (!passSelection(cellID))
+        continue;
+
+      energyMap[cellID] = hit.getEnergy();
+      hitMap.try_emplace(cellID, hit);
+    }
+  }
+
+  // ------------------------------------------------------------------
+  // Step 2: For each above-threshold crystal, check seeding conditions.
+  // ------------------------------------------------------------------
+
+  auto isLocalMax = [&energyMap,this](const uint64_t cellID,
+                             const float energy,
+                             const std::set<uint64_t>& nbrs) -> bool {
+    for (const uint64_t nb : nbrs) {
+      if (nb == cellID)
+        continue;
+
+      auto it = energyMap.find(nb);
+      if (it != energyMap.end() && it->second > energy) {
+        return false;
+      }
+    }
+
+    return true;
+  }; // lambda isLocalMax
+
+  struct SeedCandidate {
+    uint64_t cellID;
+    std::set<uint64_t> neighbors;
+  };
+
+  std::vector<SeedCandidate> seedsTypeA, seedsTypeB; // store cellIDs of seeds
+
+  for (const auto& [cellID, energy] : energyMap) {
+    if (!passSelection(cellID))
+        continue;
+
+    // --- Type A ---
+    // the seed crystal exceeds the threshold
+    // and is a local maximum within its Von Neumann neighbourhood
+    if (energy >= thrA) {
+      const auto nbrs = vonNeumannNeighbors(cellID, m_vnDistance.value());
+
+      if (isLocalMax(cellID, energy, nbrs))
+        seedsTypeA.push_back({cellID, nbrs}); // store the whole neighborhood as a seed
+    } // end Type A
+
+    // --- Type B ---
+    // the seed crystal exceeds the threshold
+    // in the same VN neighbourhood, all above-threshold neighbours are connected
+    // and the seed is a local maximum within the neighbourhood
+    if (energy >= thrB) {
+      const auto nbrs = vonNeumannNeighbors(cellID, m_vnDistance.value());
+
+      // Collect above-threshold neighbours (including self)
+      std::set<uint64_t> aboveThresh;
+      for (const uint64_t nb : nbrs) {
+        auto it = energyMap.find(nb);
+        if (it != energyMap.end() && it->second >= thrB) {
+          aboveThresh.insert(nb);
+        }
+      } // loop over neighbours
+
+      // check above threshold neighbour count
+      // and connectivity of above-threshold neighbours
+      if (aboveThresh.size() >= m_minAboveThresholdNeighbours.value() && isConnected(aboveThresh)) {
+        // Local maximum
+        if (isLocalMax(cellID, energy, nbrs))
+          seedsTypeB.push_back({cellID, nbrs}); // store the whole neighborhood as a seed
+      } // end if neighbor count and connectivity
+    } // end Type B
+  } // loop over energyMap
+
+  // ------------------------------------------------------------------
+  // Step 3: Create output clusters for the seeds
+  for (const auto& seed : seedsTypeA) {
+    if (seed.neighbors.empty())
+      continue;
+
+    auto cluster = seedsA.create();
+    cluster.setType(static_cast<int>(ClusterSeeding::SeedType::CaloDrivenA)); // Type A seed
+    // use the seed hit position as the cluster position
+    cluster.setPosition(hitMap[seed.cellID].getPosition());
+    // attach only hits that have actual energy deposits
+    for (const auto& cellID : seed.neighbors) {
+      auto it = hitMap.find(cellID);
+      if (it != hitMap.end())
+        cluster.addToHits(it->second);
+    }
+  } // loop over seedsTypeA
+
+  for (const auto& seed : seedsTypeB) {
+    if (seed.neighbors.empty())
+      continue;
+
+    auto cluster = seedsB.create();
+    cluster.setType(static_cast<int>(ClusterSeeding::SeedType::CaloDrivenB)); // Type B seed
+    // use the seed hit position as the cluster position
+    cluster.setPosition(hitMap[seed.cellID].getPosition());
+    // attach only hits that have actual energy deposits
+    for (const auto& cellID : seed.neighbors) {
+      auto it = hitMap.find(cellID);
+      if (it != hitMap.end())
+        cluster.addToHits(it->second);
+    }
+  } // loop over seedsTypeB
+
+  debug() << "CaloDrivenClusterSeeding: found " << seedsTypeA.size()
+          << " Type-A seeds and " << seedsTypeB.size() << " Type-B seeds." << endmsg;
+
+  return std::make_tuple(std::move(seedsA), std::move(seedsB));
+} // operator()
+
+DECLARE_COMPONENT(CaloDrivenClusterSeeding)
--- a/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
+++ b/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
@@ -8,31 +8,26 @@
 
 #include "ClusterSeedMerging.h" // for enum
 
+#include <cstdint>
+#include <queue>
 #include <unordered_map>
 #include <vector>
-#include <queue>
-#include <cstdint>
 
 // ============================================================
 //  CaloDrivenClusterSeeding
 // ============================================================
 
-CaloDrivenClusterSeeding::CaloDrivenClusterSeeding(const std::string& name,
-                                                   ISvcLocator*      svcLoc)
-    : ClusterSeedingBase(name, svcLoc,
-                       {KeyValues("InputCaloHitCollections", {})},
-                       {KeyValue("OutputSeedsA", "CaloDrivenSeedsA"),
-                        KeyValue("OutputSeedsB", "CaloDrivenSeedsB")}) {}
-
-StatusCode CaloDrivenClusterSeeding::initialize() {
-  return ClusterSeedingBase::initialize();
+CaloDrivenClusterSeeding::CaloDrivenClusterSeeding(const std::string& name, ISvcLocator* svcLoc)
+    : ClusterSeedingBase(name, svcLoc, {KeyValues("InputCaloHitCollections", {})},
+                         {KeyValue("OutputSeedsA", "CaloDrivenSeedsA"), KeyValue("OutputSeedsB", "CaloDrivenSeedsB")}) {
 }
+
+StatusCode CaloDrivenClusterSeeding::initialize() { return ClusterSeedingBase::initialize(); }
 
 // ------------------------------------------------------------
 
 std::tuple<edm4hep::ClusterCollection, edm4hep::ClusterCollection>
-CaloDrivenClusterSeeding::operator()(
-    const std::vector<const edm4hep::CalorimeterHitCollection*>& caloHitColls) const {
+CaloDrivenClusterSeeding::operator()(const std::vector<const edm4hep::CalorimeterHitCollection*>& caloHitColls) const {
 
   edm4hep::ClusterCollection seedsA;
   edm4hep::ClusterCollection seedsB;
@@ -68,9 +63,8 @@ CaloDrivenClusterSeeding::operator()(
   // Step 2: For each above-threshold crystal, check seeding conditions.
   // ------------------------------------------------------------------
 
-  auto isLocalMax = [&energyMap,this](const uint64_t cellID,
-                             const float energy,
-                             const std::set<uint64_t>& nbrs) -> bool {
+  auto isLocalMax = [&energyMap, this](const uint64_t cellID, const float energy,
+                                       const std::set<uint64_t>& nbrs) -> bool {
     for (const uint64_t nb : nbrs) {
       if (nb == cellID)
         continue;
@@ -93,7 +87,7 @@ CaloDrivenClusterSeeding::operator()(
 
   for (const auto& [cellID, energy] : energyMap) {
     if (!passSelection(cellID))
-        continue;
+      continue;
 
     // --- Type A ---
     // the seed crystal exceeds the threshold
@@ -165,8 +159,8 @@ CaloDrivenClusterSeeding::operator()(
     }
   } // loop over seedsTypeB
 
-  debug() << "CaloDrivenClusterSeeding: found " << seedsTypeA.size()
-          << " Type-A seeds and " << seedsTypeB.size() << " Type-B seeds." << endmsg;
+  debug() << "CaloDrivenClusterSeeding: found " << seedsTypeA.size() << " Type-A seeds and " << seedsTypeB.size()
+          << " Type-B seeds." << endmsg;
 
   return std::make_tuple(std::move(seedsA), std::move(seedsB));
 } // operator()
--- a/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
+++ b/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
@@ -31,8 +31,8 @@
 #include "edm4hep/CalorimeterHitCollection.h"
 #include "edm4hep/ClusterCollection.h"
 
-#include "k4FWCore/Transformer.h"
 #include "Gaudi/Property.h"
+#include "k4FWCore/Transformer.h"
 
 #include "ClusterSeedingBase.h"
 
@@ -40,9 +40,8 @@
 #include <vector>
 
 struct CaloDrivenClusterSeeding final
-    : ClusterSeedingBase<
-          std::tuple<edm4hep::ClusterCollection, edm4hep::ClusterCollection>(
-              const std::vector<const edm4hep::CalorimeterHitCollection*>&)> {
+    : ClusterSeedingBase<std::tuple<edm4hep::ClusterCollection, edm4hep::ClusterCollection>(
+          const std::vector<const edm4hep::CalorimeterHitCollection*>&)> {
 
   CaloDrivenClusterSeeding(const std::string& name, ISvcLocator* svcLoc);
 
@@ -63,7 +62,7 @@ struct CaloDrivenClusterSeeding final
       this, "MinAboveThresholdNeighbours", 2,
       "Minimum number of above-threshold neighbours (including seed) for a Type-B seed"};
   Gaudi::Property<int> m_vnDistance{this, "VonNeumannDistance", 2,
-      "Von Neumann distance for neighbourhood definition (applies to both seed types)"};
+                                    "Von Neumann distance for neighbourhood definition (applies to both seed types)"};
 };
 
 #endif // CaloDrivenClusterSeeding_h
--- a/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
+++ b/RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
@@ -1,0 +1,69 @@
+#ifndef CaloDrivenClusterSeeding_h
+#define CaloDrivenClusterSeeding_h 1
+
+/*
+ * CaloDrivenClusterSeeding
+ *
+ * Gaudi MultiTransformer that identifies ECAL cluster seeds purely from
+ * calorimeter hit energy deposits (no tracking information).
+ *
+ * Input
+ * -----
+ *   std::vector<const edm4hep::CalorimeterHitCollection*>
+ *       All digitised calorimeter hit collections.  The algorithm filters
+ *       internally for (ECAL) depth-0 hits
+ *
+ * Output  std::tuple<seedsA, seedsB>
+ * ------
+ *   seedsA  (edm4hep::ClusterCollection)
+ *       Type-A seeds: crystals with E > SeedEnergyThresholdA that are a
+ *       local maximum within their Von Neumann distance-k neighbourhood.
+ *
+ *   seedsB  (edm4hep::ClusterCollection)
+ *       Type-B seeds: crystals with E > SeedEnergyThresholdB that are a
+ *       local maximum within their Von Neumann distance-k neighbourhood,
+ *       have at least MinAboveThresholdNeighbours (including self) above the
+ *       above-threshold neighbours in that neighbourhood, AND all
+ *       above-threshold crystals in the neighbourhood are topologically
+ *       connected via VN-d1 adjacency.
+ */
+
+#include "edm4hep/CalorimeterHitCollection.h"
+#include "edm4hep/ClusterCollection.h"
+
+#include "k4FWCore/Transformer.h"
+#include "Gaudi/Property.h"
+
+#include "ClusterSeedingBase.h"
+
+#include <tuple>
+#include <vector>
+
+struct CaloDrivenClusterSeeding final
+    : ClusterSeedingBase<
+          std::tuple<edm4hep::ClusterCollection, edm4hep::ClusterCollection>(
+              const std::vector<const edm4hep::CalorimeterHitCollection*>&)> {
+
+  CaloDrivenClusterSeeding(const std::string& name, ISvcLocator* svcLoc);
+
+  StatusCode initialize() override;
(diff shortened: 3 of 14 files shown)


RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Steering wires the DR/SCEPCal SDs; use $K4GEO's copy if present, else fetch it.
+STEERING="$K4GEO/example/SteeringFile_IDEA_o2_v01.py"
+if [ ! -f "$STEERING" ]; then
+  [ -d k4geo ] || { git clone --no-checkout https://github.com/key4hep/k4geo && ( cd k4geo && git checkout HEAD example ); }
+  STEERING="k4geo/example/SteeringFile_IDEA_o2_v01.py"
+fi
+
+# IDEA_O2_CI=1 selects the CI wedge in both the steering and the reco config.
+export IDEA_O2_CI=1
```
jmcarcell: I think this needs to be updated after the PR in k4geo? Otherwise it's clearer if we prefer to pass arguments rather than change behaviour when exporting environment variables.

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!coll)
+      continue;
+
+    for (const auto& hit : *coll) {
+      const uint64_t cellID = hit.getCellID();
+
+      if (!passSelection(cellID))
+        continue;
+
+      energyMap[cellID] = hit.getEnergy();
```
jmcarcell, line 57: InputCaloHitCollections permits more than one collection, but this overwrites energyMap[cellID] for every hit while hitMap.try_emplace retains the first hit. With duplicate cell IDs, the seed threshold/local-maximum test uses the last energy while the output cluster attaches the first hit, so the result depends on input ordering. Could we accumulate the energy here (as TrackDrivenClusterSeeding does) and keep a defined representative hit?

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
```diff
@@ -0,0 +1,605 @@
...
+  return StatusCode::SUCCESS;
+}
+
+// ------------------------------------------------------------
+
+auto ClusterSeedGrower::connectedSeeds(const Hitmap& pool, unsigned int vnDist, float threshold,
+                                       unsigned int minHits) const -> std::vector<Hitmap> {
+  // visited tracks which cells in *pool* have already been consumed
+  // into a component.  Only cells present in *pool* are ever visited.
+  std::unordered_map<uint64_t, bool> visited;
```
jmcarcell, line 63: Nit: visited only represents membership, so std::unordered_set<uint64_t> would express the intent more directly. It would also avoid allocating a bool and pre-populating an entry for every pool cell; use pool.contains(nb) for the current pool-membership test and visited.insert(cur).second when consuming a queue entry.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
```diff
@@ -0,0 +1,33 @@
...
+
+k4run "$SOURCE_DIR/IDEA_o2_v01_reco.py"
+
+# Sanity: at least one topo cluster must have been grown.
+python3 - <<'EOF'
+import sys, podio.reading
+r = podio.reading.get_reader("testIDEA_o2_v01_reco.root")
+n = sum(len(f.get("TopoGrownClusters")) for f in r.get("events"))
+print("TopoGrownClusters produced:", n)
+sys.exit(0 if n > 0 else 1)
```
jmcarcell, line 32: The test currently only requires TopoGrownClusters to be non-empty. It would still pass if CreateOpticalCaloCells ignored the calibration constant or produced incorrect masked truth links, which are key behaviors added here. Could we add a small deterministic assertion for calibrated energy and the expected truth relation?

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    const float dotij = ni.x * nj.x + ni.y * nj.y + ni.z * nj.z;
+    if (dotij <= 0.f)
+      return false; // opposite hemisphere - never adjacent
+
+    // opening angle between the two position vectors
+    const float cosAlpha = dotij / std::sqrt(r2i * r2j);
+    const float alpha = std::acos(std::max(-1.f, std::min(1.f, cosAlpha)));
+
+    return alpha < std::atan2(mergeDist, std::sqrt(r2i));
```
jbeirer, line 107: Could we make this adjacency test symmetric? The angular cut currently only uses `r_i`, so for seeds at different radii `adjacent(i, j)` and `adjacent(j, i)` can give different results. Since this is then used to build connected components with BFS, the resulting merge can depend on the input/traversal order. Maybe use a symmetric distance criterion here (e.g. the actual spatial/chord distance, or a symmetric choice of radius)?

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
```diff
@@ -0,0 +1,326 @@
...
+        // Test i ⊆ j
+        bool isSubset = true;
+        for (const uint64_t cid : compCellIDs[i]) {
+          if (!compCellIDs[j].count(cid)) {
+            isSubset = false;
+            break;
+          }
+        }
+
+        if (isSubset) {
```
jbeirer, line 207: Does this absorption need to preserve the one-Type-C-per-component rule from the previous step? If both `i` and `j` contain a Type-C seed, a subset relation here will merge them again, effectively undoing the C+C protection in the BFS. I think we should skip the absorption when both components already contain a track-driven seed.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
```diff
@@ -0,0 +1,605 @@
...
+  // ------------------------------------------------------------------
+  // Step 2: Count total clusters; initialise per-cluster data.
+  // ------------------------------------------------------------------
+  int nClusters = 0;
+  for (const auto* coll : seedColls) {
+    if (coll)
+      nClusters += static_cast<int>(coll->size());
+  }
+
+  if (nClusters == 0) {
```
jbeirer, line 442: Should we still run the unseeded-clustering path when there are no input seeds? This early return means an event with zero A/B/C seeds produces no clusters at all, even if `hitPool` contains a group satisfying `UnseededThreshold` and `MinUnseededHits`. It seems like in this case we could just skip the seeded growth and let all of `hitPool` become the unowned pool for Step 4.

RecFCCeeCalorimeter/src/components/TrackDrivenClusterSeeding.cpp
```diff
@@ -0,0 +1,178 @@
...
+    // Add the seed cell itself. m_segmentation->neighbours() returns only the
+    // surrounding cells, so bestCell (the local-max that defines the seed) is NOT in
+    // bestNbrs and would otherwise be dropped -- leaving isolated MIP seeds with 0 hits.
+    cluster.addToHits(hitMap[bestCell]);
+
+    for (const auto& id : bestNbrs) {
+      if (id == bestCell)
+        continue; // guard against double-add if neighbours() ever includes the query cell
+      auto it = hitMap.find(id);
+      if (it != hitMap.end())
```
jbeirer, line 167: Should we also apply `SeedEnergyThreshold` to the neighbours added here? Both the class documentation and the comment above say that the Type-C seed contains the above-threshold VN-d1 neighbourhood, but `bestNbrs` is currently only filtered by the cell-ID selection, so any neighbouring hit is added regardless of energy.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
```diff
@@ -0,0 +1,262 @@
...
+    AttachIsolatedHits=False,
+    OutputLevel=INFO,
+)
+
+# ---- Truth links ----
+from Configurables import CreateTruthLinks
+
+createTruthLinksECAL = CreateTruthLinks(
+    "CreateTruthLinksECAL",
+    cell_hit_links=["SCEPCal_scint_link"],
```
jbeirer, line 216: If both scintillation and Cherenkov hits are intended to remain in `TopoGrownClusters`, should `SCEPCal_cheren_link` also be passed here? `CreateTruthLinks` uses all cluster-hit energies in the denominator, but with only the scintillation links the Cherenkov hits never contribute to any MC-particle numerator. In that configuration the cluster truth weights won't be properly normalized.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
```diff
@@ -0,0 +1,262 @@
...
+    OutputMergedSeeds="MergedSeeds",
+    MergeDistance=45.0,
+    W0=4.6,
+    OutputLevel=INFO,
+)
+
+topoClusterGrower = ClusterSeedGrower(
+    "TopoClusterGrower",
+    InputSeeds=["MergedSeeds"],
+    InputHits=["SCEPCal_digi_cheren", "SCEPCal_digi_scint"],
```
jbeirer, line 192: Is it intentional that both calibrated scintillation and Cherenkov hits contribute directly to `Cluster::energy` here? `CreateOpticalCaloCells` converts each channel independently to a cell energy, while for SCEPCal they are two measurements of the same crystal deposit. Since `ClusterSeedGrower` sets the cluster energy to the sum of all attached hit energies, including both collections here seems to effectively add the S and C energy estimates together. Should one channel drive the cluster energy, or should the two signals be kept separate and combined with the intended dual-readout energy reconstruction instead?

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
```diff
@@ -0,0 +1,605 @@
...
+        continue;
+
+      for (const auto& seed : *coll) {
+        seedTypes[ci] = seed.getType();
+        for (const auto& hit : seed.getHits()) {
+          const uint64_t cid = hit.getCellID();
+          if (!initAssigned.insert(cid).second)
+            continue; // shared hit: first seed wins
+                      // should have been resolved by ClusterSeedMerging upstream (if used)
+          initClusterHits[ci].emplace(cid, hit);
```
jbeirer, line 467: Should the initial seed hits be restricted to `hitPool` here? `hitPool` already applies `passSelection` and `HardThreshold`, but the seed contents are copied directly, so a hit below `HardThreshold` (or otherwise excluded from the pool) can still end up in the final cluster if it was part of the input seed. This seems to violate the documented meaning of `HardThreshold` as an absolute cutoff.

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
```diff
@@ -0,0 +1,168 @@
...
+  } // loop over energyMap
+
+  // ------------------------------------------------------------------
+  // Step 3: Create output clusters for the seeds
+  for (const auto& seed : seedsTypeA) {
+    if (seed.neighbors.empty())
+      continue;
+
+    auto cluster = seedsA.create();
+    cluster.setType(static_cast<int>(ClusterSeeding::SeedType::CaloDrivenA)); // Type A seed
```
jbeirer, line 135: Should this use `ClusterSeeding::encodeType(0, SeedType::CaloDrivenA)` instead? `ClusterSeedMerging.h` defines bits 0–15 of `Cluster::type` as the PDG ID and bits 16–31 as the seed-type mask, but the seed algorithms currently write the raw `SeedType` value into the lower bits. The merger also reads/ORs those raw values, so the current implementation doesn't seem to follow the declared layout and the seed information would conflict with downstream PID.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
```diff
@@ -0,0 +1,605 @@
...
+        auto it = reps.begin();
+        int winner = findRep(*it++);
+        // If multiple representatives claim this cell, merge their clusters.
+        for (; it != reps.end(); ++it)
+          mergeReps(winner, findRep(*it));
+
+        const int r = findRep(winner);
+        assigned.insert(cid);
+        clusters[r].emplace(cid, hit);
+        newFrontier[r].insert(cid);
```
jbeirer, line 287: I think `newFrontier` can lose cells when the union-find representative changes later in the same layer. For example, a hit may first be stored in `newFrontier[2]` after merging 2+3, then a later candidate merges 1+2 so the final representative becomes 1. The propagation below only looks up `newFrontier[findRep(i)]`, so the cells stored under the old representative 2 are no longer propagated and BFS can stop growing from them. Could we normalize/merge the `newFrontier` keys using the final representatives after all unions in this layer?

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.h
```diff
@@ -0,0 +1,111 @@
...
+  if (d < -pi)
+    d += 2.f * pi;
+  return d;
+}
+
+// Angular distance sqrt(dTheta^2 + dPhi_wrapped^2).
+inline float angularDist(float th1, float ph1, float th2, float ph2) {
+  const float dth = th1 - th2;
+  const float dph = deltaPhi(ph1, ph2);
+  return std::sqrt(dth * dth + dph * dph);
```
jbeirer, line 89: Should this use the actual opening angle between the two directions instead? `sqrt(dTheta^2 + dPhi^2)` is only a good angular-distance approximation around `theta = pi/2`; towards the forward region the phi term should scale with `sin(theta)`. As used for the track-seed `TrackWindow`, the current metric can therefore reject calorimeter hits that are physically well inside the requested angular cone.

RecFCCeeCalorimeter/src/components/TrackDrivenClusterSeeding.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  for (const auto* coll : caloHitColls) {
+    if (!coll)
+      continue;
+    for (const auto& hit : *coll) {
+      const uint64_t cellID = hit.getCellID();
+      if (!passSelection(cellID))
+        continue;
+      energyMap[cellID] += hit.getEnergy();
+      hitMap.try_emplace(cellID, hit);
```
jbeirer, line 57: Could we keep the emitted hit consistent with the accumulated `energyMap` here? If the same cell ID appears in more than one input collection, the seed threshold/local-maximum test uses the summed energy, while `hitMap.try_emplace` keeps only the first hit and that unsummed energy is what gets attached to the output seed. This means the energy used to create the seed can differ from the energy seen by the downstream merging/growing steps.

RecFCCeeCalorimeter/src/components/ClusterSeedingBase.h
```diff
@@ -0,0 +1,323 @@
...
+  // Returns false if:
+  //   for each (field, value) in FieldStringsToFilter:  decoder[field] != value
+  //   (i.e. the cell is filtered out)
+  bool passSelection(uint64_t cellID) const {
+    if (!m_decoder)
+      return true;
+
+    for (size_t i = 0; i < m_fieldStringsToFilter.size(); ++i) {
+      const int64_t value = m_decoder->get(cellID, m_fieldStringsToFilter[i]);
+      if (value != static_cast<int64_t>(m_fieldValuesToFilter[i]))
```
jbeirer, line 147: Could we make the naming/documentation and implementation consistent here? `FieldStringsToFilter` is documented below as excluding cells matching the configured `(field, value)` pairs, but `passSelection()` currently does the opposite: it rejects non-matching cells and therefore keeps only the configured values. The current `depth=0` config seems to rely on the latter behavior, so maybe these properties should be named/documented as a selection/include filter rather than an exclusion filter.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
```diff
@@ -0,0 +1,326 @@
...
+    while (!q.empty()) {
+      const int cur = q.front();
+      q.pop();
+      if (visited[cur])
+        continue;
+
+      // Enforce at-most-one-C rule
+      if (nodes[cur].type == static_cast<int>(ClusterSeeding::SeedType::TrackDrivenC)) {
+        if (compHasC)
+          continue; // defer this C node to its own component
```
jbeirer, line 137: I think the one-Type-C constraint still makes the result depend on traversal order here. For example, with a connected `C1 - A - C2` topology, `A` gets permanently assigned to whichever Type-C node the BFS encounters first, while the other C is deferred and later sees `A` as already visited. Swapping the Type-C input order can therefore change the merged seeds. Could shared A/B nodes be assigned to the C component using a geometric criterion instead of first-come BFS order?

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
```diff
@@ -0,0 +1,262 @@
+from Gaudi.Configuration import *
+import os
+
+# CI mode -> barrel wedge (IDEA_o2_v01_CI); otherwise the full detector.
+CI = bool(os.environ.get("IDEA_O2_CI"))
```
jbeirer, line 5: Could we parse the value of `IDEA_O2_CI` explicitly here? `bool(os.environ.get(...))` also evaluates `"0"` and `"false"` to `True`, so setting `IDEA_O2_CI=0` would unexpectedly select the CI wedge and the reduced reconstruction configuration. Something like `os.environ.get("IDEA_O2_CI") == "1"` would avoid that.

RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  // Digitize the optical hits and link each cell to its truth deposit
+  for (const auto& opticalHit : opticalHits) {
+    if (opticalHit.getEnergy() <= 0) {
+      continue; // skip cells with no signal
+    }
+    auto caloHit = outputHits.create();
+    caloHit.setCellID(opticalHit.getCellID());
+    caloHit.setPosition(opticalHit.getPosition());
+    caloHit.setEnergy(opticalHit.getEnergy() * m_calibConst.value());
```
jbeirer, line 76: Should we also set the `CalorimeterHit::type` here? `CreatePositionedCaloCells` encodes the calorimeter ID/layout/layer in this field, and downstream code such as `CaloTowerToolFCCee` relies on it to identify ECAL vs HCAL. These optical hits currently keep the default `type=0`, which would decode to an invalid calorimeter ID downstream.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
```diff
@@ -0,0 +1,33 @@
...
+#   ddsim pions into the CI wedge -> optical digi + seed/grow clustering -> check clusters.
+set -e
+
+SOURCE_DIR=$(dirname "$0")
+COMPACT="$K4GEO/FCCee/IDEA/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml"
+
+# Steering wires the DR/SCEPCal SDs; use $K4GEO's copy if present, else fetch it.
+STEERING="$K4GEO/example/SteeringFile_IDEA_o2_v01.py"
+if [ ! -f "$STEERING" ]; then
+  [ -d k4geo ] || { git clone --no-checkout https://github.com/key4hep/k4geo && ( cd k4geo && git checkout HEAD example ); }
```
jbeirer, line 12: Could we avoid falling back to an unpinned `k4geo` HEAD here? In this path the geometry still comes from the installed `$K4GEO`, but the steering file comes from whatever happens to be current on the `k4geo` default branch, so the same k4RecCalorimeter commit can change behaviour over time or mix incompatible k4geo versions. It would be better to use a steering file from the same k4geo revision as the test environment, or otherwise pin the fallback revision explicitly.

RecFCCeeCalorimeter/src/components/ClusterSeedingBase.h
```diff
@@ -0,0 +1,323 @@
...
+    float totW = 0.f;
+    for (const auto& h : hits) {
+      const float w = std::max(0.f, m_w0.value() + std::log(h.getEnergy() / totE));
+      totW += w;
+      const auto& p = h.getPosition();
+      s.x += w * p.x;
+      s.y += w * p.y;
+      s.z += w * p.z;
+    }
+    if (totW > 0.f) {
```
jbeirer, line 243: Could we add a fallback for the `totW == 0` case here? With `W0=4.6`, if no individual hit carries more than about 1% of the cluster energy all log weights are truncated to zero, and the function returns `(0,0,0)` as the barycenter despite having positive energy. That position is later used for opening-angle comparisons, so a broad valid cluster can end up with an invalid direction. Falling back to an energy-weighted barycenter would avoid this.

RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
```diff
@@ -0,0 +1,90 @@
...
+                       {KeyValue("OutputCollection", "CalorimeterHitCollection"),
+                        KeyValue("OutputLinks", "CaloHitSimCaloHitLinkCollection")}) {}
+
+StatusCode CreateOpticalCaloCells::initialize() {
+  info() << name() << ": calibration constant = " << m_calibConst.value() << endmsg;
+
+  if (m_maskCherenkov.value()) {
+    // Locate the 'cherenkov' field in the input cellID encoding to build its mask
+    const std::string opticalKey = inputLocations("OpticalHits")[0];
+    auto encoding = k4FWCore::getCellIDEncoding(opticalKey, this);
```
jbeirer, line 31: Should we also propagate the `CellIDEncoding` from `OpticalHits` to `OutputCollection` here? The output hits keep the same cell IDs, but currently the output collection doesn't get the corresponding encoding metadata. `CreateCaloCells` explicitly copies this metadata from input to output, and without it downstream components that rely on the collection metadata won't be able to decode these cell IDs. This should probably happen regardless of `maskCherenkovForTruthLink`.

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
```diff
@@ -0,0 +1,68 @@
...
+  Gaudi::Property<float> m_seedEnergyThresholdA{
+      this, "SeedEnergyThresholdA", 0.040f,
+      "Minimum crystal energy [GeV] for a Type-A seed (local max, Von Neumann neighbourhood)"};
+  Gaudi::Property<float> m_seedEnergyThresholdB{
+      this, "SeedEnergyThresholdB", 0.020f,
+      "Minimum crystal energy [GeV] for a Type-B seed (local max + connectivity)"};
+  Gaudi::Property<unsigned int> m_minAboveThresholdNeighbours{
+      this, "MinAboveThresholdNeighbours", 2,
+      "Minimum number of above-threshold neighbours (including seed) for a Type-B seed"};
+  Gaudi::Property<int> m_vnDistance{this, "VonNeumannDistance", 2,
```
jbeirer, line 64: Could this be an `unsigned int` property like the VN-distance settings in `ClusterSeedGrower`, or otherwise be checked to be non-negative? `vonNeumannNeighbors()` takes an `unsigned int`, so a negative `VonNeumannDistance` here would silently convert to a very large positive value and could make the neighbourhood BFS effectively hang.

jbeirer (changes requested):

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
SanghyunKo: Indeed I didn't presume any case of duplicated calo hits with the same cell ID across different collections. Given the assumption that there should be no duplicated cell IDs, I think the simplest solution is throwing an error on the encounter. I applied this throw to `TrackDrivenClusterSeeding` as well.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
SanghyunKo: Fixed, now it is `std::unordered_set` as suggested.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
SanghyunKo: Now the assertion also tests truth-linking sanity, though I didn't put the calibration part here (I think the calibration should be tested with more stats, e.g. in `k4DetectorPerformance`).

RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
SanghyunKo: These hits aren't produced by `CreatePositionedCaloCells` and aren't consumed by `CaloTowerToolFCCee`, and the packed `calotype + 10*caloid + 1000*layout + 10000*layer` encoding has no useful meaning for a dual-readout crystal calorimeter.

More fundamentally, I think ECAL/HCAL distinction belongs to the cell ID, where the system field already carries it unambiguously. I consider using `CalorimeterHit::type` for detector identity is a remnant of legacy code, so I'd rather not propagate that into a new detector.

RecFCCeeCalorimeter/src/components/TrackDrivenClusterSeeding.cpp
SanghyunKo: Indeed I didn't presume any case of duplicated calo hits with the same cell ID across different collections. Given the assumption that there should be no duplicated cell IDs, I think the simplest solution is throwing an error on the encounter. I applied this throw to `CaloDrivenClusterSeeding` as well (where Juan pointed out the same issue).

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
SanghyunKo: Indeed the documentation was misleading. `HardThreshold` should be a cut on the growth pool, not for all cells (we don't want to delete the input seeds, e.g. soft calo hits that are matched to the track). Fixed the doc.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.h
SanghyunKo: I changed the angular estimator from the `dR` to the opening angle (now it uses already existing `openingAngleDist` function).

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
SanghyunKo: You're right that both channels contribute, so the sum comes out at roughly twice for EM particles in the DR calo.

This is rather intended because I tend to design `ClusterSeedGrower` more detector-agnostic. The energy sum is still the correct estimator for non-DR calorimeters, while the estimator differs by particle type for DR calorimeters and must be decided in the downstream, e.g. photon & electron can use `(S+C)/2` or `S`, hadron uses `(S-chi*C)/(1-chi)`, and muon uses `S-C` for PID. I also didn't want to bake the DR correction at the clustering stage, which will make the algorithm couple to DR calibration constant, not just per-channel calibration constants.

So `Cluster::energy` here is a plain topological sum, not a calibrated energy. I modified the header to say so explicitly. `TopoGrownClusters` is an input for the downstream PF algorithm, not intended to be used out of the box.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
SanghyunKo: Indeed, now it also passes `SCEPCal_cheren_link`.

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.cpp
SanghyunKo: You're right. I shifted bits by 16 later and forgot to apply the patch everywhere. Now it is fixed.

RecFCCeeCalorimeter/src/components/CaloDrivenClusterSeeding.h
SanghyunKo: Fixed.

RecFCCeeCalorimeter/src/components/ClusterSeedingBase.h
SanghyunKo: Now it uses an energy-weighted barycenter as a fallback, but in principle that really shouldn't happen (meaning that `W0` is not properly configured). So I also inserted the warning message there.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
SanghyunKo: Yeap. Fixed that.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
```diff
@@ -0,0 +1,67 @@
...
+ddsim --compactFile="$COMPACT" --steeringFile="$STEERING" \
+      --runType batch -G --numberOfEvents 5 \
+      --gun.particle pi- --gun.energy 10*GeV \
+      --gun.direction "0.966,0.259,0.01" --random.seed 1988301045 \
+      --outputFile testIDEA_o2_v01_sim.root
+
+k4run "$SOURCE_DIR/IDEA_o2_v01_reco.py"
+
+# Sanity: digitization, truth links and clustering.
+python3 - <<'EOF'
```
jmcarcell: Can this be in its own file?

RecCalorimeter/src/components/CreateOpticalCaloCells.cpp
SanghyunKo: Indeed, now we propagate the cell ID encoding.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
SanghyunKo: Now we allow running the unseeded clustering path when there is no seed is found. But since this may indicate wrong configuration in the upstream seeding stage, I put a warning message there.

RecFCCeeCalorimeter/src/components/ClusterSeedingBase.h
SanghyunKo: Modified the documentation so that the code and the inline comments are more consistent.

RecFCCeeCalorimeter/src/components/TrackDrivenClusterSeeding.cpp
SanghyunKo: Here the energy threshold should be only applied to the seed crystal, but not to the others (otherwise the neighboring cell will never survive). I made this clear in the inline comment.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
SanghyunKo: Now in the release build, it detects the k4geo's tag and checkout the right version. The nightly build still checkouts the `HEAD` of the `main` branch.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: Now we use `min(r_i, r_j)` to make it symmetric (though it wouldn't impact very much in practice).

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: Now we don't merge two track-driven seeds in the downstream absorption path as well. To prevent this fix from producing empty clusters in the conflict resolve stage (every cell conflicts in a subset by definition), I added the rule that the track-driven cluster should never lose it's seed cell (where the track landed on the ECAL surface).

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
SanghyunKo: Thanks for spotting this, now the `newFrontier` is re-keyed with the final representatives before propagation, folding together entries whose stale keys resolve to the same root.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: Indeed. Rather than `continue` at the first encounter, now the rule runs after the BFS: a component with several track seeds becomes one per track seed, each grows outward along the adjacency chain without entering another, and the seeds it reaches go to the nearest in angle.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
SanghyunKo: Now the bash heredoc became an own python script.

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
jmcarcell: Maybe either `example` should be installed in k4geo or the steering file moved somewhere where it is installed

SanghyunKo: Perhaps we can make a new folder like `config` or `steering` in k4geo and install? I see that the current `example` is not really well-organized.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
jbeirer: Looks good, thanks. The comment above still describes the old r_i-only criterion though, would be good if that can be adapted

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
jbeirer: The C+C guard looks good, but I think if a track-containing component is absorbed into a non-track component, the `anchorCell` isn't propagated, so the protection added below can be lost. What about

```cpp
if (components[i].hasTrackSeed) {
	components[j].hasTrackSeed = true;
	components[j].anchorCell = components[i].anchorCell;
}
```

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
```diff
@@ -0,0 +1,262 @@
...
+    OutputLevel=INFO,
+)
+
+topoClusterGrower = ClusterSeedGrower(
+    "TopoClusterGrower",
+    InputSeeds=["MergedSeeds"],
+    InputHits=["SCEPCal_digi_cheren", "SCEPCal_digi_scint"],
+    OutputClusters="TopoGrownClusters",
+    ReadoutName="SCEPCal_Main",
+    FieldStringsToFilter=[],
```
jbeirer: Should we use the companion-hit mechanism here, i.e. grow only on `cherenkov=0` and add `cherenkov=1` afterwards? Otherwise the S and C signals from the same crystal are grown independently and can end up in different clusters, which would change the per-cluster S/C ratio.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
```diff
@@ -0,0 +1,610 @@
...
+      rep[b] = a;
+  };
+
+  // --- Helper shared by both strategy branches ---
+  // Returns all cells in pool that neighbour fid (within vnDist), are not yet
+  // assigned to any cluster, and are above growThreshold.
+  auto eligibleNeighbors = [&](uint64_t fid) {
+    std::vector<uint64_t> result;
+    for (const uint64_t nb : vonNeumannNeighbors(fid, vnDist)) {
+      if (nb == fid || assigned.count(nb))
```
jbeirer: For `MergeClusters`, should already assigned neighbours still be checked for ownership? Right now two unseeded clusters only merge if they claim the same unassigned cell in the same layer; if one front reaches a cell already owned by the other, they stay separate even though the grown region is connected.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
```diff
@@ -0,0 +1,399 @@
...
+      std::unordered_set<int> reached;
+      std::queue<int> frontier;
+      frontier.push(trackNodes[k]);
+
+      while (!frontier.empty()) {
+        const int cur = frontier.front();
+        frontier.pop();
+
+        for (const int nb : comp) {
+          if (nb == cur || isTrackSeed(nb) || !adjacent(cur, nb))
```
jbeirer: Would it make sense to build the adjacency lists once and reuse them here? For a component with several track seeds this recomputes `adjacent()` for every track traversal, so the splitting can scale like O(T*N²) with fairly expensive geometric calculations.

jbeirer (changes requested):

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: The comment is now also fixed.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: Fixed.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
SanghyunKo: I deliberately picked this approach because I found out that using only scintillation channel often fails to grow the energy deposit from MIP-like particles.

RecFCCeeCalorimeter/src/components/ClusterSeedGrower.cpp
SanghyunKo: I think that'd be extremely rare case for the default VN-2 setup (every possible topological path should be a even distance), but it could be possible for VN-1. Fixed.

RecFCCeeCalorimeter/src/components/ClusterSeedMerging.cpp
SanghyunKo: IMO the clustering was never bottleneck (it was usually GEANT4 simulation and ML inference downstream), but it's true that we can do better here. Fixed.

RecFCCeeCalorimeter/tests/options/IDEA_o2_v01_reco.py
jbeirer: I see, thanks for clarifying. Have you checked whether S and C from the same crystal ending up in different clusters has any noticeable effect on the reconstructed energy or PID, or is this already handled downstream?

SanghyunKo: The artifact that you concern happens at the boundary of the cluster, which carries very little energy. I ran 1k events of Z→qq events (43k clusters, 83 TeV clustered) to quantify the effect:

Basically, S and C of the same crystal land in different clusters for 0.43% of crystals, and the energy that actually changes cluster is 0.092% of the clustered total.

For the per-cluster S/C ratio: 93.5% of clusters are untouched, and among the affected ones the median shift is 1.7%. For clusters above 5 GeV — the ones driving energy and PID — it's 0.35%.

(so the overall effect is really negligible)

RecFCCeeCalorimeter/tests/options/test_IDEAo2_reco.sh
jbeirer: I think the cleanest solution would be to install the IDEA steering file(s) into `share/k4geo/example/`, so this test can just use `$K4GEO/example/SteeringFile_IDEA_o2_v01.py` without any cloning/version-detection logic. I’d be fine with handling that as a follow-up in k4geo rather than blocking this PR though.

jbeirer:
Hi @SanghyunKo,

this looks good to me now, thanks for addressing all the comments.

Do you maybe have any slides or other studies validating the clustering performance that you could link in the PR description? I think it would be useful to have some reference for the physics performance/validation of the new algorithms.

The k4geo steering-file handling could probably be cleaned up in a follow-up by installing the steering file with k4geo, but I don't think that needs to block this PR.

jbeirer (approved):

SanghyunKo:
@jbeirer Sure, I added the link to my talk in the PR description.

jbeirer:
I think the Ubuntu26 failure happens because the tests load the stack's older k4RecCalorimeter plugins instead of the ones built from this PR, I think https://github.com/HEP-FCC/k4RecCalorimeter/pull/283 should fix this

jmcarcell:
This is ready to go, time for last comments or :+1: for merging

PR #272, opened 2026-07-06
Configure indexing service in test scripts.

scott-snyder:
Update test scripts to configure indexing service where it will be needed.

BEGINRELEASENOTES
- Update test scripts to configure indexing service where it will be needed.
ENDRELEASENOTES


giovannimarchiori (approved):
Thanks @scott-snyder , approved

PR #274, opened 2026-07-07
Add GATr-based Pi0 identification

jacofan:
BEGINRELEASENOTES

- Add a functional (`TRAPPISTPi0PhotonInference.cpp`) to produce a classification score for a cluster running inference using a pre-trained ONNX model based on a GATr architecture. The score is added to cluster's shapeParameters and a new cluster collection is produced.
- Add a functional (`ClusterPi0PhotonID.cpp`) to identify pi0s and build two new `RecoParticle` collections from a cluster comparing the classification score and a user-defined threshold.

ENDRELEASENOTES


RecFCCeeCalorimeter/src/components/Pi0IDTool.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "edm4hep/Vector3f.h"
+
+// STL
+#include <optional>
+#include <string>
+#include <vector>
+
+/** @struct Pi0IDTool
+ *
+ * Gaudi MultiTransformer that identifies pi0 particles from a collection of clusters
```
jmcarcell: Maybe should not say it's a tool not to confuse it with tools from Gaudi. It may also be a good idea to change the name of the algorithm not to have Tool in it.

RecFCCeeCalorimeter/src/components/Pi0IDTool.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  Gaudi::Property<float> m_threshold{this, "Threshold", 0.5,
+                                     "Threshold for particle identification based on the shape parameter value"};
+
+  edm4hep::Vector3f calculateMomentum(double energy, const edm4hep::Vector3f& position,
+                                      const edm4hep::Vector3f& origin) const {
+    double dirx = position.x - origin.x;
+    double diry = position.y - origin.y;
+    double dirz = position.z - origin.z;
+    double quadsum_dir = sqrt(pow(dirx, 2) + pow(diry, 2) + pow(dirz, 2));
```
jmcarcell:
```suggestion
    double quadsum_dir = std::sqrt(dirx * dirx + diry * diry + dirz * dirz);
```
Prefer `std::` math functions, some like `abs` (without `std::`) don't do what you would expect. Also no need to call another function to just do the square.

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,116 @@
...
+    debug() << "Input cluster has " << shapeParameterNames.size() << " names in metadata" << endmsg;
+    // Find and save the index of the shape parameter to be used
+    auto it = std::find(shapeParameterNames.begin(), shapeParameterNames.end(), m_shapeParameterScoreName.value());
+    if (it != shapeParameterNames.end()) {
+      m_scoreIndex = std::distance(shapeParameterNames.begin(), it);
+      debug() << "Feature " << m_shapeParameterScoreName.value() << " found in position " << m_scoreIndex.value()
+              << " of collection metadata" << endmsg;
+    } else {
+      error() << "Feature " << m_shapeParameterScoreName.value() << " not found, aborting..." << endmsg;
+      throw std::runtime_error("Feature " + m_shapeParameterScoreName.value() + " not found in collection metadata");
```
jmcarcell:
```suggestion
      throw std::runtime_error("Feature " + m_shapeParameterScoreName.value() + " not found in collection metadata");
```
Don't print the error message twice

RecFCCeeCalorimeter/src/components/TRAPPISTPi0PhotonInference.cpp
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  std::tuple<edm4hep::ClusterCollection> operator()(const edm4hep::ClusterCollection& unpairedClusters) const override {
+    edm4hep::ClusterCollection outputClusters;
+    if (unpairedClusters.empty()) {
+      debug() << "Input cluster collection is empty. No clusters to process" << endmsg;
+      return std::make_tuple(std::move(outputClusters));
+    } else {
+      debug() << "Processing " << unpairedClusters.size() << " unpaired clusters" << endmsg;
+    }
```
jmcarcell:
```suggestion
    debug() << "Processing " << unpairedClusters.size() << " unpaired clusters" << endmsg;
```
No strong opinion on this, but since it will return with an empty output cluster collection if the input is empty, you can just print how many you are processing and make it a bit shorter. It's the same information.

RecFCCeeCalorimeter/src/components/TRAPPISTTool.cpp
```diff
@@ -0,0 +1,151 @@
...
+  // ONNX session: manages the loaded model and performs inference
+  std::unique_ptr<Ort::Session> m_ortSession{nullptr};
+  // Input and output names for the ONNX model, used to identify tensors during inference
+  std::vector<const char*> m_input_names;
+  std::vector<const char*> m_output_names;
+  // Struct to hold input data
+  struct GATrInput {
+    std::vector<float> positions;
+    std::vector<float> energies;
+  };
```
jmcarcell: I'm not sure you win much by using this. In the end you are calling `clusterData.positions` all the time; if you had two different vectors named `positions` or `inputPositions` it would be the same.

RecFCCeeCalorimeter/src/components/Pi0IDTool.cpp
jacofan: Suggestion applied. The names now describe the algorithms more clearly

RecFCCeeCalorimeter/src/components/Pi0IDTool.cpp
jacofan: Done

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
jacofan: Suggestion applied

RecFCCeeCalorimeter/src/components/TRAPPISTPi0PhotonInference.cpp
jacofan: True, suggestion applied

RecFCCeeCalorimeter/src/components/TRAPPISTTool.cpp
jacofan: Suggestion applied

RecFCCeeCalorimeter/tests/options/ALLEGRO_o1_v03.sh
```diff
@@ -32,6 +32,7 @@ if ! test -f ./cellNoise_map_endcapTurbine_electronicsNoiseLevel.root; then  # a
   download_file "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/neighbours_map_ecalE_turbine.root"
   download_file "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/neighbours_map_ecalB_thetamodulemerged_ecalE_turbine_hcalB_hcalEndcap_phitheta.root"
   download_file "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/cellNoise_map_endcapTurbine_electronicsNoiseLevel.root"
+  download_file "https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/ALLEGRO/ALLEGRO_o1_v03/TRAPPIST_topoclustering.onnx"
```
jmcarcell, line 35: This means the model can be changed and results will change, making future results possibly unreproducible. In how we use models for tests in other places, like in k4RecTracker, at least they are hashed.

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    return StatusCode::SUCCESS;
+  }
+
+  std::tuple<edm4hep::ReconstructedParticleCollection, edm4hep::ReconstructedParticleCollection>
+  operator()(const edm4hep::ClusterCollection& clustersWithScore) const override {
+    edm4hep::ReconstructedParticleCollection outputPi0Particles;
+    edm4hep::ReconstructedParticleCollection outputPhotonParticles;
```
SanghyunKo: Maybe already discussed in the meeting; why do we need to emit two `ReconstructedParticleCollection`? Can't it be just one collection with different PDG of the `ReconstructedParticle`?

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,132 @@
...
+    return std::make_tuple(std::move(outputPi0Particles), std::move(outputPhotonParticles));
+  }
+
+  StatusCode finalize() override { return StatusCode::SUCCESS; }
+
+private:
+  std::optional<std::size_t> m_scoreIndex = std::nullopt;
+
+  Gaudi::Property<std::string> m_shapeParameterScoreName{
+      this, "ShapeParameterName", "ClusterTRAPPISTScore",
```
SanghyunKo: I'd rename the shape parameter name more self-explanatory.

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    particle.addToClusters(cluster);
+
+    const double energy = cluster.getEnergy();
+    particle.setEnergy(energy);
+
+    particle.setPDG(pdg);
+    particle.setCharge(0);
+    particle.setMass(mass);
+    particle.setGoodnessOfPID(pidScore);
```
SanghyunKo, line 145: Personally, I'd prefer emitting a separate `edm4hep::ParticleID` collection and linking to the corresponding `edm4hep::ReconstructedParticle`. The issue is that this algorithm is not the only one who wants to occupy the `goodnessOfPID` variable. Suppose the following case:

For a `ReconstructedParticle` that is identified as pi0, storing pi0 ID score is an obvious choice. However, for a `ReconstructedParticle` that is identified as photons, there could be multiple possible cases for the choice of `goodnessOfPID`, e.g. photon ID score.

(honestly I think it makes more sense either to make `goodnessOfPID` a vector member or completely delegate the PID to `edm4hep::ParticleID`, but this is not the scope of this PR)

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,132 @@
...
+    double pz = energy * dirz / quadsum_dir;
+    return edm4hep::Vector3f(px, py, pz);
+  }
+
+  void buildParticle(edm4hep::ReconstructedParticleCollection& collection, const edm4hep::Cluster& cluster, int pdg,
+                     float mass, float pidScore) const {
+
+    auto particle = collection.create();
+
+    particle.addToClusters(cluster);
```
SanghyunKo, line 137: I'm not entirely sure about the cluster linking strategy. If we link the `ReconstructedParticle` to the clusters with score, then both the previous cluster collection (which is used by other algorithms) and the cluster with embedded pi0 PID score should be persistent in the file, despite they carry completely duplicated information besides the PID score, which is stored in the `ReconstructedParticle` anyway (or in `edm4hep::ParticleID` as I suggested below).

I'd recommend to link it to the previous cluster collection (so that we can safely remove the cluster with PID without breaking the links), or update the link for all other algorithms to use the cluster with PID.

At least with the current approach, the linked cluster from the new `ReconstructedParticle` collection and the linked clusters from other algorithms won't resolve despite they are exactly the same physical cluster.

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
```diff
@@ -0,0 +1,132 @@
...
+  operator()(const edm4hep::ClusterCollection& clustersWithScore) const override {
+    edm4hep::ReconstructedParticleCollection outputPi0Particles;
+    edm4hep::ReconstructedParticleCollection outputPhotonParticles;
+    for (const auto& cluster : clustersWithScore) {
+      // Retrieve the score from the shape parameters and if it exceeds the threshold
+      // create and fill a new ReconstructedParticle
+      float clusterScore = cluster.getShapeParameters(m_scoreIndex.value());
+      debug() << "Cluster score: " << clusterScore << ", threshold: " << m_threshold << endmsg;
+      if (clusterScore > m_threshold) {
+        buildParticle(outputPi0Particles, cluster, 111, 0.1349768, clusterScore);
```
SanghyunKo: Can we make the mass of the pi0 configurable? For instance, I'd avoid the situation that pi0s from Pandora and this algorithm carry tiny numerical differences for their mass.

RecFCCeeCalorimeter/tests/options/ALLEGRO_o1_v03.sh
jacofan: After discussing with @giovannimarchiori and @BrieucF we agreed to stick with the current approach for now and postpone any comprehensive refactoring of the tests to another PR

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
BrieucF: For analyzer convenience it is easier IMHO to have mono-particle collections. But it is right that if we want to align with e.g. what CLD does, we should output one big reco particle collection.

Let me discuss this with Giovanni as well

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
BrieucF: I think this should be handled at the steering file level, we can call the inference right after producing the clusters, drop the collection without the TRAPPIST score  and make sure everything else that links cluster does it with the augmented collection. I would feel it unnatural to link to the clusters without the score

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
BrieucF: Yes, let's add a ParticleID object which is the standard way to address this.

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
BrieucF: Ok, let's go for one recoParticle collection and leave the selection based on PID to later anal;ysis/n-tuple making. Being aligned with other detectors/algorithm is important

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
jacofan: Done

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
jacofan: Thanks for the feedback, done

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
jacofan: Done

RecFCCeeCalorimeter/src/components/ClusterPi0PhotonID.cpp
jacofan: Done

PR #275, opened 2026-07-21
Fix phi wrapping in topo cluster barycenter calculation.

scott-snyder:
Take phi wrapping into account when calculating the barycenter and width of topo clusters.

In the process, adapt some code from ATLAS to help with phi calculations.

BEGINRELEASENOTES
- Topo clustering wasn't taking phi wrapping into account, meaning that clusters very near +-pi could have catastrophically wrong phi positions and widths. 
ENDRELEASENOTES

### DIFF
--- a/RecCaloCommon/CMakeLists.txt
+++ b/RecCaloCommon/CMakeLists.txt
@@ -3,6 +3,7 @@
 ################################################################################
 
 find_package(boost_timer REQUIRED)
+find_package(boost_unit_test_framework REQUIRED)
 file(GLOB _sources src/*.cpp )
 
 gaudi_add_library(RecCaloCommon
@@ -46,4 +47,11 @@ if(BUILD_TESTING)
     LINK DD4hep::DDCore k4FWCore::k4Interface k4FWCore::k4FWCore
     TEST)
   target_include_directories(MultiIndexer_test.exe AFTER PUBLIC include)
+
+
+  gaudi_add_executable(phihelper_test.exe
+    SOURCES tests/phihelper_test.cpp
+    LINK Boost::unit_test_framework
+    TEST)
+  target_include_directories(phihelper_test.exe AFTER PUBLIC include)
 endif()
--- a/RecCaloCommon/include/RecCaloCommon/phihelper.h
+++ b/RecCaloCommon/include/RecCaloCommon/phihelper.h
@@ -22,26 +22,44 @@ namespace k4::recCalo {
  * Wrap angle in radians to [-pi, pi]
  *
  * Odd positive (negative) multiples of pi map to (-)pi.
+ * (May not always be true for the long double case.)
+ *
+ * If phi is too large (larger than the largest integer that can be
+ * represented exactly as a T), this may fail to produce a value
+ * in the specified range.
  */
 template <std::floating_point T>
-inline T wrapToPi(T phi) {
-  constexpr T PI = std::numbers::pi_v<T>;
-  // For large values this is faster:
-  if (phi < -100 || phi > 100) {
-    return std::remainder(phi, 2 * PI);
+inline constexpr T wrapToPi(T phi) {
+  constexpr T TWOPI = 2 * std::numbers::pi_v<T>;
+  constexpr T INV2PI = std::numbers::inv_pi_v<T> / 2;
+  T x = phi * INV2PI;
+
+  // Round x to the nearest integer.
+  // https://stackoverflow.com/questions/17035464/a-fast-method-to-round-a-double-to-a-32-bit-int-explained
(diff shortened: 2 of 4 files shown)


RecCaloCommon/include/RecCaloCommon/phihelper.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace k4::recCalo {
+
+/**
+ * Wrap angle in radians to [-pi, pi]
+ *
+ * Odd positive (negative) multiples of pi map to (-)pi.
+ */
+template <std::floating_point T>
+inline T wrapToPi(T phi) {
```
jmcarcell: `std::reminder(phi, 2 * PI)` already gives you the value wrapped in [-pi, pi] so no function is needed? (and certainly not decrease or increase by 2*PI in a loop until the right value is reached...)

RecCaloCommon/CMakeLists.txt
```diff
@@ -46,4 +47,11 @@ if(BUILD_TESTING)
     LINK DD4hep::DDCore k4FWCore::k4Interface k4FWCore::k4FWCore
     TEST)
   target_include_directories(MultiIndexer_test.exe AFTER PUBLIC include)
+
+
+  gaudi_add_executable(phihelper_test.exe
```
jmcarcell, line 52: All these executables are created and never run on CI so there is no guarantee (over time) any of them work

RecCaloCommon/include/RecCaloCommon/phihelper.h
jmcarcell: If speed is the reason, then using a division seems to be fastest:

Image

scott-snyder: hi -

Thanks for the comments.  This may be getting rather far away from the fairway,
but...

The reduction function is almost always called with an argument in the
range -2pi..2pi, or at least not much more than then, so that's what's
important for performance.  With that constraint, iteratively adding 2pi
is probably not a bad way to do this, and that's traditionally how this
sort of thing has been done.

I think when this was made into a more general utility, the std::remainder
case was added to keep things from falling over in the (rare) case that
it gets called with something larger.

So i think what we have is not bad, though the arbitrary threshold
of 100 could perhaps be lowered.  But it could be improved...

I tried to do some timing.  I get results at least qualitatively similar
to you.

First of all, with gcc16 / x86_64, i see std::remainder compiling to an fprem
instruction.  That's probably why that's relatively slow.

You didn't give your exact code for division + floor (and it's not obvious
if that would be rounding correctly).  I tried to implement
something like this:

```
  constexpr T TWOPI = 2*std::numbers::pi_v<T>;
  constexpr T inv2pi = std::numbers::inv_pi_v<T>/2;
  T x = phi * inv2pi;
  T ix = std::round(x);
  return phi - TWOPI*ix;
```

Similar to what you showed, this is basically constant over the input range.
However, for small inputs (|phi|<pi), it is 2-3 times slower than the
original version (and std::remainder is about 8 times slower).

But std::round compiles to a call to the roundfNN library function, which
is what's slowing that down.  We can do the rounding instead with the 0x18
trick:

```
  constexpr T TWOPI = 2*std::numbers::pi_v<T>;
  constexpr T inv2pi = std::numbers::inv_pi_v<T>/2;
  T x = phi * inv2pi;
  constexpr T toint = 0x1.8p0 * (1ul<<(std::numeric_limits<T>::digits-1));
  T ix = (x + toint) - toint;
  return phi - TWOPI*ix;
```

and in that case, this is about 10% slower than the original for
|phi|<pi and faster for larger values.

But this isn't entirely what we want, since the function is specified as

```
   * Odd positive (negative) multiples of pi map to (-)pi.
```

The rounding trick above gives halves rounding to even integers (banker's
rounding), while we want halves to round to zero.  I couldn't think of
a nice way to do it, so i tried like this:

```
  constexpr T TWOPI = 2*std::numbers::pi_v<T>;
  constexpr T inv2pi = std::numbers::inv_pi_v<T>/2;
  T x = phi * inv2pi;
  constexpr T toint = 0x1.8p0 * (1ul<<(std::numeric_limits<T>::digits-1));
  T ix = (x + toint) - toint;
  T diff = ix - x;
  if (std::abs(diff) == 0.5) {
    if (ix > 0) -- ix;
    else if (ix < 0) ++ix;
  }
  return phi - TWOPI*ix;
```

which makes no measureable difference in timing and gets the rounding
correct (except for the long double case, which i don't think we really
care about).

While i was at it, i also added constexpr to all these functions; why not.

RecCaloCommon/CMakeLists.txt
scott-snyder: The TEST keyword adds them to the tests.  I know ctest runs them; i just verified that they run in the github CI.

PR #276, opened 2026-07-28
Fix SW clustering sensitivity to cell ordering.

scott-snyder:
The results of SW clustering depend on the order in which cells are listed in the input container, due to FP non-associativity in sums over cells. These sums are done in single-precision, so we can fix this in practice by doing the sums in double precision and only rounding to single precision at the end.

BEGINRELEASENOTES
- Fix SW clustering to avoid depending on the order in which cells appear in the input container.
ENDRELEASENOTES


giovannimarchiori (approved):

giovannimarchiori (approved):
looks good to me, approved

PR #277, opened 2026-07-29
Fix cluster<>mcparticle links.

scott-snyder:
When trying to match cluster cells with reconstructed cells, need to look at cell id, not id().  The latter is just the index in the container, and we're looking at completely different containers here.  (The collectionID doesn't save us since it's not available for containers that were read from a file.)

BEGINRELEASENOTES
- Fix bug in cluster <-> MC particle association.
ENDRELEASENOTES


giovannimarchiori (approved):
bug fix, approved

